PR #1, opened 2019-07-15
linint2

khallock:
Add linint2 function

### DIFF
--- a/src/geocat/comp/__init__.py
+++ b/src/geocat/comp/__init__.py
@@ -1,4 +1,5 @@
 from . import _ncomp
+from .version import __version__
 import numpy as np
 import xarray as xr
 import dask.array as da
@@ -129,7 +130,6 @@ def linint2(fi, xo, yo, icycx, msg=None, meta=True, xi=None, yi=None):
                 specified as a keyword argument.
 
     Returns:
-
         :class:`xarray.DataArray`: The interpolated grid. If the *meta*
         parameter is True, then the result will include named dimensions
         matching the input array. The returned value will have the same
--- a/src/geocat/comp/_ncomp/ncomp.pyx
+++ b/src/geocat/comp/_ncomp/ncomp.pyx
@@ -125,7 +125,112 @@ cdef set_ncomp_msg(ncomp.ncomp_missing* ncomp_msg, num):
     elif ncomp_type == ncomp.NCOMP_LONGDOUBLE:
         ncomp_msg.msg_longdouble = ncomp_to_dtype[ncomp_type](num)
 
+@cython.embedsignature(True)
+@cython.binding(True)
 def _linint2(np.ndarray xi, np.ndarray yi, np.ndarray fi, np.ndarray xo, np.ndarray yo, int icycx, msg=None):
+    """Interpolates a regular grid to a rectilinear one using bi-linear
+    interpolation.
+
+    linint2 uses bilinear interpolation to interpolate from one
+    rectilinear grid to another. The input grid may be cyclic in the x
+    direction. The interpolation is first performed in the x direction,
+    and then in the y direction.
+
+    Args:
+
+        xi (:class:`numpy.ndarray`):
+            An array that specifies the X coordinates of the fi array.
+            Most frequently, this is a 1D strictly monotonically
+            increasing array that may be unequally spaced. In some
+            cases, xi can be a multi-dimensional array (see next
+            paragraph). The rightmost dimension (call it nxi) must have
+            at least two elements, and is the last (fastest varying)
+            dimension of fi.
+
+            If xi is a multi-dimensional array, then each nxi subsection
+            of xi must be strictly monotonically increasing, but may be
+            unequally spaced. All but its rightmost dimension must be
+            the same size as all but fi's rightmost two dimensions.
+
+            For geo-referenced data, xi is generally the longitude
+            array.
+
+        yi (:class:`numpy.ndarray`):
+            An array that specifies the Y coordinates of the fi array.
+            Most frequently, this is a 1D strictly monotonically
+            increasing array that may be unequally spaced. In some
+            cases, yi can be a multi-dimensional array (see next
+            paragraph). The rightmost dimension (call it nyi) must have
+            at least two elements, and is the second-to-last dimension
+            of fi.
+
+            If yi is a multi-dimensional array, then each nyi subsection
+            of yi must be strictly monotonically increasing, but may be
+            unequally spaced. All but its rightmost dimension must be
+            the same size as all but fi's rightmost two dimensions.
+
+            For geo-referenced data, yi is generally the latitude array.
+
+        fi (:class:`numpy.ndarray`):
+            An array of two or more dimensions. If xi is passed in as an
+            argument, then the size of the rightmost dimension of fi
+            must match the rightmost dimension of xi. Similarly, if yi
+            is passed in as an argument, then the size of the second-
+            rightmost dimension of fi must match the rightmost dimension
+            of yi.
+
+            If missing values are present, then linint2 will perform the
+            bilinear interpolation at all points possible, but will
+            return missing values at coordinates which could not be
+            used.
+
+        xo (:class:`numpy.ndarray`):
+            A one-dimensional array that specifies the X coordinates of
+            the return array. It must be strictly monotonically
+            increasing, but may be unequally spaced.
+
+            For geo-referenced data, xo is generally the longitude
+            array.
+
+            If the output coordinates (xo) are outside those of the
+            input coordinates (xi), then the fo values at those
+            coordinates will be set to missing (i.e. no extrapolation is
+            performed).
+
+        yo (:class:`numpy.ndarray`):
+            A one-dimensional array that specifies the Y coordinates of
+            the return array. It must be strictly monotonically
+            increasing, but may be unequally spaced.
+
+            For geo-referenced data, yo is generally the latitude array.
+
+            If the output coordinates (yo) are outside those of the
+            input coordinates (yi), then the fo values at those
+            coordinates will be set to missing (i.e. no extrapolation is
+            performed).
+
+        icycx (:obj:`bool`):
+            An option to indicate whether the rightmost dimension of fi
+            is cyclic. This should be set to True only if you have
+            global data, but your longitude values don't quite wrap all
+            the way around the globe. For example, if your longitude
+            values go from, say, -179.75 to 179.75, or 0.5 to 359.5,
+            then you would set this to True.
+
+        msg (:obj:`numpy.number`):
+            A numpy scalar value that represent a missing value in fi.
+            This argument allows a user to use a missing value scheme
+            other than NaN or masked arrays, similar to what NCL allows.
+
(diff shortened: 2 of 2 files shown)


src/geocat/comp/__init__.py
```diff
@@ -3,7 +3,115 @@
...
-class Ncomp(object):
-    def __init__(self, xarray_obj):
-        self._obj = xarray_obj
+def linint2(fi, xo, yo, icycx, xmsg=None, meta=False, xi=None, yi=None):
+    """Return a 2-dimensionally linearly interpolated grid.
+
+    Args:
+
+        fi (:class:`xarray.DataArray` or :class:`numpy.ndarray`):
+            Geopotential height in [m] with the rightmost dimensions being
```
clyne: Is this really "Geopotential height in [m]..."? Isn't it simply an input array with two or more dimensions?

src/geocat/comp/__init__.py
```diff
@@ -3,7 +3,115 @@
 import xarray as xr
 from dask.array.core import map_blocks
 
-@xr.register_dataarray_accessor('ncomp')
-class Ncomp(object):
-    def __init__(self, xarray_obj):
-        self._obj = xarray_obj
+def linint2(fi, xo, yo, icycx, xmsg=None, meta=False, xi=None, yi=None):
+    """Return a 2-dimensionally linearly interpolated grid.
```
clyne: Perhaps a more accurate synopsis rather than:

"Return a 2-dimensionally linearly interpolated grid."

 is to say that this function

"interpolates a regular grid to a rectilinear one using bi-linear interpolation. "

src/geocat/comp/__init__.py
```diff
@@ -3,7 +3,115 @@
...
+
+    if xmsg is None:
+        xmsg = _ncomp.dtype_default_fill[fi.dtype]
+
+    if xi is None:
+        xi = fi.coords[fi.dims[-1]].values
+    if yi is None:
+        yi = fi.coords[fi.dims[-2]].values
+    fi_data = fi.data
+    fo_chunks = list(fi.chunks)
```
clyne: Per our discussion the other day my understanding is that because the  input interpolation dimensions (two fastest varying dimensions) do not match the output interpolation dimensions the data cannot be blocked along the interpolation dimensions. What will happen if the user provides an input array that *is* blocked along the interpolation dimensions ?

src/geocat/comp/__init__.py
```diff
@@ -3,7 +3,115 @@
...
+    fo_chunks[-2:] = (yo.shape, xo.shape)
+    fo = map_blocks(_ncomp._linint2, xi, yi, fi_data, xo, yo, icycx, xmsg, chunks=fo_chunks, dtype=fi.dtype, drop_axis=[fi.ndim-2, fi.ndim-1], new_axis=[fi.ndim-2, fi.ndim-1])
```
clyne: what does map_blocks do? Need a comment

src/geocat/comp/__init__.py
khallock: It is **not** supposed to be "Geopotential height in [m]...", thanks for catching that @clyne!

I originally copied this docstring from the WRF-Python function `slp` to use as a template and apparently forgot to change the description for this argument. Your description is correct, I'll submit an additional commit with an updated docstring shortly.

src/geocat/comp/__init__.py
```diff
@@ -3,7 +3,115 @@
...
+        fi (:class:`xarray.DataArray` or :class:`numpy.ndarray`):
+            Geopotential height in [m] with the rightmost dimensions being
+            bottom_top x south_north x west_east.
+
+            Note:
+
+                This variable must be
+                supplied as a :class:`xarray.DataArray` in order to copy the
+                dimension names to the output.  Otherwise, default names will
+                be used.
```
clyne: How are missing data values handled? I.e. how does the user indicate that the data may contain missing values? 

More generally, we seem to be missing some of the function description from the NCL documentation:

linint2 uses bilinear interpolation to interpolate from one rectilinear grid to another. The input grid may be cyclic in the x direction. The interpolation is first performed in the x direction, and then in the y direction.

If missing values are present, then linint2 will perform the bilinear interpolation at all points possible, but will return missing values at coordinates which could not be used.

If the output coordinates (xo,yo) are outside those of the input coordinates (xi,yi), then the fo values at those coordinates will be set to missing (i.e. no extrapolation is performed).

src/geocat/comp/__init__.py
khallock: As of commit d853943 there is a comment describing what `map_blocks` does here.

src/geocat/comp/__init__.py
khallock: Fixed in commit 21cbc18

src/geocat/comp/__init__.py
khallock: In commit 28c5d94 I added a check to ensure that the chunked shapes of the interpolation (fastest varying) dimensions of the input array always match the full shape of those dimensions. Existing chunking across any additional dimensions will remain intact.

src/geocat/comp/__init__.py
khallock: Missing values are a piece of the puzzle that hasn't been *fully* solved yet, although I believe the current behavior is acceptable for the moment (NaN missing values work implicitly as long as "-ffast-math" isn't used to compile Fortran in NComp).

At the top of the software stack we've got Xarray, which uses NaNs to represent missing values. Because of this, it seems unlikely that a user would ever actually need to specify a missing value for an `xarray.DataArray`.

The Cython layer of our software acts as a translation from NumPy-like arrays (`numpy.ndarray` or `dask.array.Array`) to the `ncomp_array` structs used in NComp's C API. While NumPy does allow NaN values in its arrays, the preferred method of representing missing values is by using `numpy.ma.MaskedArray`, which itself has an actual data array and a boolean mask of equal sizes.

NCL supports missing values by attaching a `_FillValue` attribute to an array, which indicates that any element in that array that matches `_FillValue` should be considered a missing element.

The compiled code in NComp expects arrays to contain missing values using NCL's missing value scheme. Unfortunately, this means that at some point between the high-level Python (Xarray) interface and the lowest level Fortran code in NComp we will have to translate between missing value implementations.

I'm currently working on a solution (one more graceful than simply assuming NaNs will be skipped over in Fortran) that can handle `xarray.DataArray`s with NaNs, but not `numpy.ma.MaskedArray`s.

src/geocat/comp/__init__.py
clyne: Is it reasonable (good behavior) to re-chunk the users input array? This seems like an unintended side effect of using linint. Maybe we should just throw an error...

src/geocat/comp/__init__.py
clyne: There are definitely tradeoffs between using NaNs to indicate missing data and using Numpy's masked arrays. However, Xarray's missing data support is built around NaNs. If you think masked arrays are a better option we should probably have some discussion about this and how we would maintain compatibility with Xarray. 

Otherwise it would seem that we just need to make sure missing values are always represented as NaNs all of the way through the software stack (since the Fortran code uses them at the bottom, and Xarray uses them at the top), AND we need to document that linint operates correctly when presented with NaNs. I.e. it treats them as missing data and does something reasonable with them during interpolation. 

I may be missing something here. If so we can talk.

src/geocat/comp/__init__.py
khallock: Re-chunking should not actually modify the chunking on the original array, but it *may* create a duplicate of the input array temporarily if the newly defined chunks are not already contiguous in memory.

Regardless, you make a good point that it might be better behavior to just throw an error. As of commit 30d85f5, I've defined a new `ChunkError` exception to be thrown when an input array in chunked in a way that is incompatible with an underlying Fortran routine, and I removed the re-chunking step and now throw a `ChunkError` instead.

clyne:
Per discussion today a couple more items are needed by @khallock 

1. Reference documentation on the web site

2. An example in the form of a Jupyter notebook

abb58:
> Per discussion today a couple more items are needed by @khallock
> 
>     1. Reference documentation on the web site
> 
>     2. An example in the form of a Jupyter notebook

Was facing some difficulty with adding/uploading .ipynb extensions on git. 
```
#!/usr/bin/env python
# coding: utf-8

import numpy as np
import xarray as xr
import geocat.comp
import cartopy.crs as ccrs
import matplotlib.pyplot as plt
import matplotlib.ticker as mticker
from matplotlib import cm
from cartopy.mpl.ticker import LongitudeFormatter, LatitudeFormatter
from cartopy.mpl.geoaxes import GeoAxes
from mpl_toolkits.axes_grid1 import AxesGrid

# Open a netCDF data file using xarray default engine and load the data into xarrays
ds = xr.open_dataset("/Users/bagusett/projects/sst.nc")
sst = ds.TEMP[0,0,:,0:-1]
lat = ds.LAT[:]
lon = ds.LON[0:-1]

# Provide interpolation grid and call $linint2$ function from $geocat-comp$
newlat = np.linspace(min(lat), max(lat), 128)
newlon = np.linspace(min(lon), max(lon), 360)
newsst = geocat.comp.linint2(sst, newlon, newlat, 0)

# Visualize the interpolated grid $newsst$ using matplotlib and cartopy
#### -----  Graphics using cartopy and matplotlib ----- ####
projection = ccrs.PlateCarree()
axes_class = (GeoAxes, dict(map_projection=projection))
fig = plt.figure(figsize=(10,8))
axgr = AxesGrid(fig, 111, axes_class=axes_class,
                nrows_ncols=(2, 1),
                axes_pad=0.7,
                cbar_location='bottom',
                cbar_mode='single',
                cbar_pad=0.1,
                cbar_size='10%',
                label_mode='')  # note the empty label_mode

for i, ax in enumerate(axgr):
    ax.coastlines()
    ax.set_xticks(np.linspace(-180, 180, 13), crs=projection)
    ax.set_yticks(np.linspace(-60, 60, 5), crs=projection)
    lon_formatter = LongitudeFormatter(zero_direction_label=True)
    lat_formatter = LatitudeFormatter()
    ax.xaxis.set_major_formatter(lon_formatter)
    ax.yaxis.set_major_formatter(lat_formatter)

    # Plot contours for both the subplots
    if( i==0 ):
        X, Y = np.meshgrid(lon, lat)
        p=ax.contourf(X, Y, sst, levels=16, transform=projection, cmap=cm.jet)
        ax.set_title('Original Grid', fontsize=14, fontweight='bold')
    else:
        newX, newY = np.meshgrid(newlon, newlat)
        p=ax.contourf(newX, newY, newsst, levels=16, transform=projection, cmap=cm.jet)
        ax.set_title('ReGrid - linint2', fontsize=14, fontweight='bold')

# Add color bar and label details (title, size, etc.)
cax=axgr.cbar_axes[0]
cax.colorbar(p)
axis=cax.axis[cax.orientation]
axis.label.set_text('Temperature ($^{\circ} C$)')
axis.label.set_size(16)
axis.major_ticklabels.set_size(10)

# Save figure and show
plt.savefig('linint2', dpi=300)
plt.show()
```
linint2

PR #32, opened 2019-09-23
rcm2rgrid, rgrid2rcm

abb58:
Complete implementation for rcm2rgrid, rgrid2rcm. An example python script for the examples and plotting functionality would be updated as a different PR.


abb58:
An example plot for rcm2rgrid:
rcm2rgrid

khallock:
I agree with @maboualidev regarding `__init__.py` becoming bloated; putting everything in this file was done as a starting point, but isn't sustainable as we add more functions.

There is certainly potential for using namespaces under `geocat.comp`, but I'd be concerned about creating excessive namespaces -- `geocat.comp.eof.eofunc()` is clear but not concise. If we're going to go down that road, I think we may want to consider ditching the `comp` namespace and create the `eof` namespace directly on `geocat`: `geocat.eof.eofunc()`.

Note that the package as a whole would still be called "GeoCAT-comp", we would just be changing the `import geocat.???` statements.

Ultimately, this discussion is entirely unrelated to this PR and is not a blocker, just thought I'd chime in with @maboualidev's comment.

clyne:
> I agree with @maboualidev regarding `__init__.py` becoming bloated; putting everything in this file was done as a starting point, but isn't sustainable as we add more functions.
> 
> There is certainly potential for using namespaces under `geocat.comp`, but I'd be concerned about creating excessive namespaces -- `geocat.comp.eof.eofunc()` is clear but not concise. If we're going to go down that road, I think we may want to consider ditching the `comp` namespace and create the `eof` namespace directly on `geocat`: `geocat.eof.eofunc()`.
> 
> Note that the package as a whole would still be called "GeoCAT-comp", we would just be changing the `import geocat.???` statements.
> 
> Ultimately, this discussion is entirely unrelated to this PR and is not a blocker, just thought I'd chime in with @maboualidev's comment.

How about open a separate refactoring issue to address this point, since it will impact all of the functions? 

@abb58  Has the plot been fixed yet?

maboualidev:
> I agree with @maboualidev regarding `__init__.py` becoming bloated; putting everything in this file was done as a starting point, but isn't sustainable as we add more functions.
> 
> There is certainly potential for using namespaces under `geocat.comp`, but I'd be concerned about creating excessive namespaces -- `geocat.comp.eof.eofunc()` is clear but not concise. If we're going to go down that road, I think we may want to consider ditching the `comp` namespace and create the `eof` namespace directly on `geocat`: `geocat.eof.eofunc()`.
> 
> Note that the package as a whole would still be called "GeoCAT-comp", we would just be changing the `import geocat.???` statements.
> 
> Ultimately, this discussion is entirely unrelated to this PR and is not a blocker, just thought I'd chime in with @maboualidev's comment.

Agree. Smaller path is better.

src/geocat/comp/__init__.py
```diff
@@ -231,3 +231,245 @@ def linint2(fi, xo, yo, icycx, msg=None, meta=True, xi=None, yi=None):
...
+
+    if lat2d.shape[0] < 2 or lon2d.shape[0] < 2 or lat2d.shape[1] < 2 or lon2d.shape[1] < 2:
+        raise Exception("ERROR rcm2rgrid: The input/output lat/lon grids must have at least 2 elements !")
+
+    if fi.ndim < 2:
+        raise Exception("ERROR rcm2rgrid: fi must be at least two dimensions !\n")
+
+    if fi.shape[fi.ndim - 2] != lat2d.shape[0] or fi.shape[fi.ndim - 1] != lon2d.shape[1]:
+        raise Exception("ERROR rcm2rgrid: The rightmost dimensions of fi must be (nlat2d x nlon2d),"
+                        "where nlat2d and nlon2d are the size of the lat2d/lon2d arrays !")
```
khallock: I think it would be best to create a new exception type to raise here similarly to in `linint2`. `Exception` is Python's base exception class, and raising it directly is potentially troublesome as it makes it very difficult to construct meaningful `try/except` statements. Further up in \_\_init\_\_.py, I created new `Error` and `ChunkError` classes; perhaps we could create a new `DimensionError` or something similar to raise in this case?

src/geocat/comp/__init__.py
```diff
@@ -231,3 +231,245 @@ def linint2(fi, xo, yo, icycx, msg=None, meta=True, xi=None, yi=None):
...
+
+    if lat2d.shape[0] < 2 or lon2d.shape[0] < 2 or lat2d.shape[1] < 2 or lon2d.shape[1] < 2:
+        raise Exception("ERROR rgrid2rcm: The input/output lat/lon grids must have at least 2 elements !")
+
+    if fi.ndim < 2:
+        raise Exception("ERROR rgrid2rcm: fi must be at least two dimensions !\n")
+
+    if fi.shape[fi.ndim - 2] != lat1d.shape[0] or fi.shape[fi.ndim - 1] != lon1d.shape[0]:
+        raise Exception("ERROR rgrid2rcm: The rightmost dimensions of fi must be (nlat1d x nlon1d),"
+                        "where nlat1d and nlon1d are the size of the lat1d/lon1d arrays !")
```
khallock: See previous comment for lines 302-312.

src/geocat/comp/__init__.py
abb58: Resolved the `Exception` with the new `Error` class

src/geocat/comp/__init__.py
abb58: Resolved the `Exception` with the new `Error` class

setup.py
```diff
@@ -15,9 +15,21 @@
...
+except KeyError:
+    PREFIX = sys.prefix
+    NCOMP_INC = os.path.join(PREFIX, 'include')
+    NCOMP_LIB = os.path.join(PREFIX, 'lib')
+
+include_dirs = [NCOMP_INC, numpy.get_include()]
+library_dirs = [NCOMP_LIB]
+
+print(include_dirs)
+print(library_dirs)
```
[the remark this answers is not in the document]
abb58: Right. This shouldn't be in the commits. I have reverted accordingly.

src/geocat/comp/__init__.py
khallock: Thanks for updating this, it's definitely better than using `raise Exception()`.

However, I think there's still an opportunity to be more verbose here. When I created these error classes, I meant for `Error` to not be raised directly but rather to be used as a base class for other GeoCAT-comp error classes, currently including `ChunkError` and `CoordinateError`. While raising `Error` directly is perfectly valid Python, I'd like it to be possible to use
```
try:
    # do something
except geocat.comp.Error:
    # catch any geocat.comp Error
```
to catch _any_ `geocat.comp` error while still being able to be more specific if desired:
```
try:
    # do something
except geocat.comp.CoordinateError:
    # catch only CoordinateError
```

I think we'll eventually have a collection of Error subclasses that can be used to cover most of our error cases, but it seems like for now something like `DimensionError` might be relevant to add in this situation.

src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -653,3 +652,222 @@ def _moc_globe_atl(np.ndarray lat_aux_grid, np.ndarray a_wvel, np.ndarray a_bolu
...
+	sometimes, there are small gaps in the interpolated grids. Any interior points not
+	interpolated in the initial interpolation pass will be filled using linear interpolation.
+        In some cases, edge points may not be filled.
+
+    """
+    cdef libncomp.ncomp_array* ncomp_lat2d = np_to_ncomp_array(lat2d)
+    cdef libncomp.ncomp_array* ncomp_lon2d = np_to_ncomp_array(lon2d)
+    cdef libncomp.ncomp_array* ncomp_fi    = np_to_ncomp_array(fi)
+    cdef libncomp.ncomp_array* ncomp_lat1d = np_to_ncomp_array(lat1d)
+    cdef libncomp.ncomp_array* ncomp_lon1d = np_to_ncomp_array(lon1d)
```
khallock: Could we update this to use the `_ncomp.Array` class if possible? I'd be glad to help out if you'd like.

src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -653,3 +652,222 @@ def _moc_globe_atl(np.ndarray lat_aux_grid, np.ndarray a_wvel, np.ndarray a_bolu
...
+	NARR (North American Regional Reanalysis) models/datasets. No extrapolation is
+	performed beyond the range of the input coordinates. The method used is simple inverse
+	distance weighting. Missing values are allowed but ignored.
+
+    """
+    cdef libncomp.ncomp_array* ncomp_lat1d = np_to_ncomp_array(lat1d)
+    cdef libncomp.ncomp_array* ncomp_lon1d = np_to_ncomp_array(lon1d)
+    cdef libncomp.ncomp_array* ncomp_fi    = np_to_ncomp_array(fi)
+    cdef libncomp.ncomp_array* ncomp_lat2d = np_to_ncomp_array(lat2d)
+    cdef libncomp.ncomp_array* ncomp_lon2d = np_to_ncomp_array(lon2d)
```
khallock: Could we update this to use the _ncomp.Array class if possible? I'd be glad to help out if you'd like.

src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -653,3 +652,222 @@ def _moc_globe_atl(np.ndarray lat_aux_grid, np.ndarray a_wvel, np.ndarray a_bolu
     # output.numpy[output.numpy == output_missing_value] = np.nan
 
     return output.numpy
+
+def _rcm2rgrid(np.ndarray lat2d, np.ndarray lon2d, np.ndarray fi, np.ndarray lat1d, np.ndarray lon1d, msg=None):
```
khallock: I think we should add the `@carrayify` decorator here to ensure the input `np.ndarray`s are contiguous in memory.

src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -653,3 +652,222 @@ def _moc_globe_atl(np.ndarray lat_aux_grid, np.ndarray a_wvel, np.ndarray a_bolu
...
+    if ncomp_fo.type == libncomp.NCOMP_DOUBLE:
+        fo_msg = ncomp_fo.msg.msg_double
+    else:
+        fo_msg = ncomp_fo.msg.msg_float
+
+    fo[fo == fo_msg] = np.nan
+
+    return fo
+
+def _rgrid2rcm(np.ndarray lat1d, np.ndarray lon1d, np.ndarray fi, np.ndarray lat2d, np.ndarray lon2d, msg=None):
```
khallock: I think we should add the `@carrayify` decorator here to ensure the input `np.ndarray`s are contiguous in memory.

abb58: Done !

src/geocat/comp/_ncomp/ncomp.pyx
abb58: Done

src/geocat/comp/__init__.py
abb58: Done! https://github.com/NCAR/geocat-comp/pull/32/commits/7737d79d0c984530262b229131280782c922402a

src/geocat/comp/__init__.py
abb58: Done ! https://github.com/NCAR/geocat-comp/pull/32/commits/7737d79d0c984530262b229131280782c922402a

src/geocat/comp/_ncomp/ncomp.pyx
abb58: Done ! https://github.com/NCAR/geocat-comp/pull/32/commits/49f85e998500ffef88a83dad66c34cd1c5ff2309

src/geocat/comp/_ncomp/ncomp.pyx
abb58: Done ! https://github.com/NCAR/geocat-comp/pull/32/commits/49f85e998500ffef88a83dad66c34cd1c5ff2309

PR #33, opened 2019-10-01
Add carrayify decorator to Cython code

khallock:
Using this decorator on a function in ncomp.pyx will ensure that all of that function's numpy.ndarray arguments are C-contiguous in memory.

The decorator function iterates over each input argument, checks if the argument is an instance of numpy.ndarray, then checks if its CARRAY flag is set (meaning the C_CONTIGUOUS, ALIGNED, and BEHAVED flags are all True), and then calls numpy.ascontiguousarray() on the argument array and inserts the new contiguous array into the argument list in place of the original non-contiguous array.

This decorator uses functools.wraps to ensure docstrings of all wrapped functions are preserved.

### DIFF
--- a/src/geocat/comp/_ncomp/ncomp.pyx
+++ b/src/geocat/comp/_ncomp/ncomp.pyx
@@ -8,6 +8,21 @@ import numpy as np
 cimport numpy as np
 import functools
 
+def carrayify(f):
+    """
+    A decorator that ensures that :class:`numpy.ndarray` arguments are
+    C-contiguous in memory. The decorator function takes no arguments.
+    """
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        new_args = list(args)
+        for i, arg in enumerate(new_args):
+            if isinstance(arg, np.ndarray) and not arg.flags.carray:
+                new_args[i] = np.ascontiguousarray(arg)
+        return f(*new_args, **kwargs)
+    return wrapper
+
+
 dtype_default_fill = {
              "DEFAULT_FILL":       ncomp.DEFAULT_FILL_DOUBLE,
              np.dtype(np.int8):    np.int8(ncomp.DEFAULT_FILL_INT8),
--- a/test/test_carrayify.py
+++ b/test/test_carrayify.py
@@ -1,0 +1,19 @@
+import numpy as np
+import xarray as xr
+from geocat.comp._ncomp import carrayify
+
+import sys
+import time
+import unittest as ut
+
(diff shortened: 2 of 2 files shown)


src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -8,6 +8,21 @@ import numpy as np
...
+def carrayify(f):
+    """
+    A decorator that ensures that :class:`numpy.ndarray` arguments are
+    C-contiguous in memory. The decorator function takes no arguments.
+    """
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        new_args = list(args)
+        for i, arg in enumerate(new_args):
+            if isinstance(arg, np.ndarray) and not arg.flags.carray:
```
abb58: Would there be a case where `arg` is not a `np.ndarray` by the time it reaches `carrify(f)`

khallock: Yes, for example:
```
def _linint2(np.ndarray xi, np.ndarray yi, np.ndarray fi, np.ndarray xo, np.ndarray yo, int icycx, msg=None):
```

`icycx` is an `int` and `msg` is a Python object.

abb58: Got it. Thanks !

test/test_carrayify.py
```diff
@@ -0,0 +1,19 @@
...
+
+class Test_carrayify(ut.TestCase):
+    def test_carrayify_one_arg(self):
+        @carrayify
+        def new_func(a):
+            return a
+        a = np.array([1, 2, 3])
+        b = a[::-1]
+        c = new_func(a)
+        self.assertTrue(a.flags.carray)
```
maboualidev: So, does this means the original input array is also modified?

if yes, are we ok with that, i.e. modifying the input array memory pattern without warning user?

khallock: No, the original input array will not be modified by `@carrayify`.

This decorator intercepts a function call (`new_func` in this test case) and iterates through its argument list. A new argument list is constructed that matches the original, except that any `numpy.ndarray` arg that does not have `arg.flags.carray = True` will be copied into a new contiguous array (using `numpy.ascontiguousarray`) in place of the original array argument in the new argument list. The new argument list is then passed to the invocation of `new_func`, so the original input array never even makes it into `new_func`.

maboualidev: Sounds good to me

PR #38, opened 2019-10-31
eofunc function added to Python

maboualidev:
This pull request adds the ``eofunc`` and ``eofunc_n`` functionality of NCL to GeoCAT-comp. Rather than providing two different functions as in NCL, both functions are provided through the same API. In case if the last dimension is not time, the user could pass ``time_dim`` optional argument to determine the time dimension.

Other than the tests that are written in ncomp to test C/C++/FORTRAN side, additional 49 tests are added:

```bash
Testing started at 10:43 AM ...
/Users/mabouali/miniconda3/envs/geocat-comp/bin/python "/Users/mabouali/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-C/ch-0/192.6817.19/PyCharm CE.app/Contents/helpers/pycharm/_jb_unittest_runner.py" --path /Users/mabouali/NCAR/GitHub/GeoCAT/geocat-comp/test/test_eofunc.py
Launching unittests with arguments python -m unittest /Users/mabouali/NCAR/GitHub/GeoCAT/geocat-comp/test/test_eofunc.py in /Users/mabouali/NCAR/GitHub/GeoCAT/geocat-comp/test



Ran 49 tests in 0.024s

OK

Process finished with exit code 0
```


src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -30,6 +30,11 @@ def carrayify(f):
     return wrapper
 
 
+class GeocatArray(object):
+    def __init__(self, data, attrs={}):
+        self.data = data
+        self.attrs = attrs
+
```
khallock: I think we implemented this as a proof-of-concept of a possible mechanism for sending multiple return values (including attributes, not just a single array) between the `geocat.comp._ncomp` Cython code and the `geocat.comp` Python code. It does not look like we are using it at this point and are instead just using tuples, but I wanted to make a note here in case we wish to revisit the idea in the future.

maboualidev: You are right. We are not using that.
I removed it from the source code.

src/geocat/comp/__init__.py
```diff
@@ -245,3 +245,88 @@ def linint2(fi, xo, yo, icycx, msg=None, meta=True, xi=None, yi=None):
...
+    attrs = {}
+    # converting the keys to string instead of bytes also fixing matrix and method
+    # TODO: once Kevin's work on char * is merged, we could remove this part or change it properly.
+    for k, v in response[1].items():
+        if k in {b'matrix', b'method'}:
+            attrs[k.decode('utf-8')] = v.tostring().decode('utf-8')[:-1]
+        else:
+            attrs[k.decode('utf-8')] = v
+
+    return eof, attrs
```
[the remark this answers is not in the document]
maboualidev: I thought you wanted everything to be ``numpy.ndarray`` and you want the attributes and data to be returned as Tuple; that's the reason I was returning a tuple.

Based on the new instructions, I change the code to return an ``xarray.DataArray`` and associated the attributes to the it.

maboualidev:
> This looks great!
> 
> Just a few comments:
> I think we may want to remove the files in `test/__pycache__` that seem to have been inadvertently included in the repo. This would be a good case to use the "squash and merge" option when merging the PR to avoid those files showing up anywhere in the git history.
> 
> We could also add a `meta=` keyword argument that mimics the `_Wrap` variants of NCL, which could copy dimension names and/or other relevant metadata from the `data` input argument.
> 
> There are a several additional comments inline with the code.

Removed the ``test/__pycache__`` folder.

Also added the support for `meta=` keyword. If the input array is of type ``xarray.DataArray`` and the ``meta`` keyword is set to something that evaluates to ``True``, all the attributes from the input data are transferred to the output.

Also, added couple of more UnitTests to test for ``meta`` keyword support.

maboualidev:
Somethings wrong with the `CircleCI` script. It could not find `conda`

```bash
Resolving deltas: 100% (1956/1956), done.
Branch 'eofunc' set up to track remote branch 'eofunc' from 'origin'.
Switched to a new branch 'eofunc'
Traceback (most recent call last):
  File "/usr/local/bin/conda", line 12, in <module>
    from conda.cli import main
ModuleNotFoundError: No module named 'conda'
Exited with code 1
```

khallock:
I just merged develop into eofunc again. There was a second fix to address the `no module named 'conda'` error we've been seeing, and with that it looks like the eofunc branch is building again now.

src/geocat/comp/__init__.py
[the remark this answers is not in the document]
khallock: I'm sorry for any confusion, I'll try to clarify what I meant.

When I said that, I meant that I think complex (as in more than just a single array) return values from the Cython interface would be better implemented as a `(array, attrs)` tuple as opposed to subclassing `numpy.ndarray` or creating our own class, and it looks like you're [already doing that in Cython](https://github.com/NCAR/geocat-comp/blob/619cd84c4d4b0f1b7c7083d108b6b7b9968831c7/src/geocat/comp/_ncomp/ncomp.pyx#L456). 👍

For our top-level Python interface, I think we should make our best effort to use `xarray.DataArray` (and `xarray.DataSet` when appropriate) as much as possible, even when a user provides NumPy arrays as input. At this point, Xarray is the de facto Python data model for working with atmospheric/climate science data and is a hard dependency of `geocat.comp` (it must be installed); as an additional benefit, using Xarray saves us from having to invent our own mechanism for passing attributes/coordinates/dimensions around.

Having a robust NumPy interface at the Cython layer means that we can simplify our Python code and also fit in well with the geosciences Python community at large by standardizing around the use of Xarray; any users that truly wish to avoid using Xarray can still access the underlying Cython/NumPy functions as desired.

I'm sorry again for the misunderstanding. I hope this helps, but please let me know if you have any other questions.

src/geocat/comp/__init__.py
```diff
@@ -326,7 +339,7 @@ def eofunc(data, neval, **kwargs):
         else:
             attrs[k.decode('utf-8')] = v
 
-    return eof, attrs
+    return xr.DataArray(response[0], attrs=attrs)
```
khallock: In addition to attaching `attrs`, we might want to also copy `data.dims` and `data.coords` from the input array when `meta=True`, but filtering out `time_dim`; something like this:
```
dims = ["evn"] + data.dims[:time_dim] + data.dims[time_dim+1:]
```

NCL has an internal `eofMeta_n` function used to copy metadata that might be a useful reference: https://github.com/NCAR/ncl/blob/develop/ni/src/examples/gsun/contributed.ncl#L6810

src/geocat/comp/__init__.py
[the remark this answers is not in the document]
maboualidev: Sounds good to me: (1) Cython Level, returns (results: np.ndarray, attrs: dict), (2) Python level returns results: (xr.DataArray, xr.DataSet)

src/geocat/comp/__init__.py
maboualidev: Made the changes based on your comment to include `dims` and `coords` in the output `xr.DataArray`.

PR #41, opened 2019-11-12
Refactor cython arrays

khallock:
Created a new _ncomp.Array class to simplify the process of interacting with a NumPy and ncomp_array that refer to the same data.

### DIFF
--- a/src/geocat/comp/_ncomp/ncomp.pxd
+++ b/src/geocat/comp/_ncomp/ncomp.pxd
@@ -1,3 +1,4 @@
+# cython: language_level=3
 cdef extern from "ncomp/constants.h":
     cdef double DEFAULT_FILL "DEFAULT_FILL_DOUBLE";
     cdef char   DEFAULT_FILL_INT8 "NC_FILL_BYTE";
@@ -77,11 +78,11 @@ cdef extern from "ncomp/wrapper.h":
 
     int eofunc(const ncomp_array * x_in, const int neval_in,
                const ncomp_attributes * options_in,
-               ncomp_array* x_out, ncomp_attributes* attrList_out) nogil;
+               ncomp_array** x_out, ncomp_attributes* attrList_out) nogil;
 
     int eofunc_n(const ncomp_array * x_in, const int neval_in,
                  const int t_dim,
                  const ncomp_attributes * options_in,
-                 ncomp_array * x_out, ncomp_attributes * attrList_out) nogil;
+                 ncomp_array ** x_out, ncomp_attributes * attrList_out) nogil;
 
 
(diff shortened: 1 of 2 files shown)


khallock:
Essentially what this PR provides is a new Cython class that represents a unified NumPy and ncomp_array.

The way the Cython code has generally worked thus far is that a NumPy array is passed in and is used to create an ncomp_array that reuses the NumPy array's data pointer (the code that does this is tucked away in the `np_to_ncomp_array` function):
```
def _linint2(np.ndarray xi,...):
    ...
    cdef libncomp.ncomp_array* ncomp_xi = np_to_ncomp_array(xi)
    # use xi for additional NumPy operations in the Cython function
    # use ncomp_xi for calling NComp C API functions
```

The NumPy array and ncomp_array exist as separate variables and there is no obvious indication that modifying the data of one of them will affect the other as well.

My new code abstracts this relationship a bit by creating an `Array` object that has both a NumPy array and ncomp_array, accessible as the attributes `.numpy` and `.ncomp`:
```
def _linint2(np.ndarray xi_np,...):
    ...
    xi = Array.from_np(xi_np)
    # use xi.numpy for additional NumPy operations in the Cython function
    # use xi.ncomp for calling NComp C API functions
```

An `Array` object can be instantiated using either the `Array.from_np` or `Array.from_ncomp` static methods; note that calling `Array()` directly will raise an exception.

Additionally, the `Array` class has a `__dealloc__` method that is called when an `Array` object is destroyed (goes out of scope). In particular, the output `ncomp_array*` structs that are allocated in NComp (and whose data eventually becomes owned by the NumPy array to be returned by the Cython function) should be freed using `ncomp_array_free` (which frees the struct itself as well as its `size_t *shape` array).

khallock:
@maboualidev Thanks for the suggestion, adding unit tests for this feature is a great idea. Unfortunately because the factory methods `from_ncomp` and `from_np` are defined using `cdef`, they're only callable from within Cython and are not callable directly from Python, so it's tricky to test them similarly to our existing unit tests.

I'm trying to create a workaround Python test script that compiles a Cython test program on demand, which can then be used to test the `cdef` functions. Hopefully I'll have something to add to this PR soon.

maboualidev:
> @maboualidev Thanks for the suggestion, adding unit tests for this feature is a great idea. Unfortunately because the factory methods `from_ncomp` and `from_np` are defined using `cdef`, they're only callable from within Cython and are not callable directly from Python, so it's tricky to test them similarly to our existing unit tests.
> 
> I'm trying to create a workaround Python test script that compiles a Cython test program on demand, which can then be used to test the `cdef` functions. Hopefully I'll have something to add to this PR soon.

Sounds great to me.

khallock:
@maboualidev I would like to merge this PR now and add the Cython unit testing as a separate PR. Did you have any concerns other than the unit tests that you wanted to address before approving this?

maboualidev:
> @maboualidev I would like to merge this PR now and add the Cython unit testing as a separate PR. Did you have any concerns other than the unit tests that you wanted to address before approving this?

Approved from my side.

PR #44, opened 2019-11-15
Moc globe atl

erogluorhan:
This pull request adds `moc_globe_atl` function of NCL into GeoCAT-comp. 

7 different test cases have been written to test if output data types, dimensions, and contents are correct for inputs with varying data types as well as user-defined or NaN missing value inputs. All test cases pass successfully.

### DIFF
--- a/src/geocat/comp/__init__.py
+++ b/src/geocat/comp/__init__.py
@@ -249,7 +249,7 @@ def linint2(fi, xo, yo, icycx, msg=None, meta=True, xi=None, yi=None):
 
 
 def moc_globe_atl(lat_aux_grid, a_wvel, a_bolus, a_submeso, tlat, rmlak,
-                  msg=None, meta=True):
+                  msg=None, meta=False):
     """Facilitates calculating the meridional overturning circulation for the
     globe and Atlantic.
 
@@ -302,10 +302,10 @@ def moc_globe_atl(lat_aux_grid, a_wvel, a_bolus, a_submeso, tlat, rmlak,
 
     # Ensure input arrays are numpy.ndarrays
     if isinstance(lat_aux_grid, xr.DataArray):
-    lat_aux_grid = lat_aux_grid.values
+        lat_aux_grid = lat_aux_grid.values
 
     if isinstance(a_wvel, xr.DataArray):
-    a_wvel = a_wvel.values
+        a_wvel = a_wvel.values
 
     if isinstance(a_bolus, xr.DataArray):
         a_bolus = a_bolus.values
@@ -319,43 +319,15 @@ def moc_globe_atl(lat_aux_grid, a_wvel, a_bolus, a_submeso, tlat, rmlak,
     if isinstance(rmlak, xr.DataArray):
         rmlak = rmlak.values
 
-    fi_data = fi.data
 
-    if isinstance(fi_data, da.Array):
-        chunks = list(fi.chunks)
-
-        # ensure rightmost dimensions of input are not chunked
-        if chunks[-2:] != [yi.shape, xi.shape]:
-            raise ChunkError("linint2: the two rightmost dimensions of fi must"
-                             " not be chunked.")
-
-        # ensure rightmost dimensions of output are not chunked
-        chunks[-2:] = (yo.shape, xo.shape)
-
-        # map_blocks maps each chunk of fi_data to a separate invocation of
-        # _ncomp._linint2. The "chunks" keyword argument should be the chunked
-        # dimensionality of the expected output; the number of chunks should
-        # match that of fi_data. Additionally, "drop_axis" and "new_axis" in
-        # this case indicate that the two rightmost dimensions of the input
-        # will be dropped from the output array, and that two new axes will be
-        # added instead.
-        fo = map_blocks(_ncomp._linint2, xi, yi, fi_data, xo, yo, icycx, msg,
-                        chunks=chunks, dtype=fi.dtype,
-                        drop_axis=[fi.ndim-2, fi.ndim-1],
-                        new_axis=[fi.ndim-2, fi.ndim-1])
-    elif isinstance(fi_data, np.ndarray):
-        fo = _ncomp._linint2(xi, yi, fi_data, xo, yo, icycx, msg)
-    else:
-        raise TypeError
+    # Call ncomp function
+    out_arr = _ncomp._moc_globe_atl(lat_aux_grid, a_wvel, a_bolus, a_submeso,
+                                    tlat, rmlak, msg)
 
     if meta:
-        coords = {k:v if k not in fi.dims[-2:]
-                  else (xo if k == fi.dims[-1] else yo)
-                  for (k, v) in fi.coords.items()}
-
-        fo = xr.DataArray(fo, attrs=fi.attrs, dims=fi.dims,
-                              coords=coords)
+        pass
+        # TODO: Retaining possible metadata might be revised in the future
     else:
-        fo = xr.DataArray(fo)
+        out_arr = xr.DataArray(out_arr)
 
-    return fo
+    return out_arr
--- a/src/geocat/comp/__init__.py
+++ b/src/geocat/comp/__init__.py
@@ -245,3 +245,117 @@ def linint2(fi, xo, yo, icycx, msg=None, meta=True, xi=None, yi=None):
         fo = xr.DataArray(fo)
 
     return fo
+
+
+
+def moc_globe_atl(lat_aux_grid, a_wvel, a_bolus, a_submeso, tlat, rmlak,
+                  msg=None, meta=True):
+    """Facilitates calculating the meridional overturning circulation for the
+    globe and Atlantic.
+
+    Args:
+
+        lat_aux_grid (:class:`numpy.ndarray`):
+            Latitude grid for transport diagnostics.
+
+        a_wvel (:class:`numpy.ndarray`):
+            Area weighted Eulerian-mean vertical velocity [TAREA*WVEL].
+
+        a_bolus (:class:`numpy.ndarray`):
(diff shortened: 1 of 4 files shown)


src/geocat/comp/__init__.py
```diff
@@ -354,4 +354,92 @@ def eofunc(data, neval, **kwargs) -> xr.DataArray:
...
+            a_submesa is of type double. Otherwise, the return type will be float.
+
+
+    Examples:
+
+        # TODO: To be included
+
+    """
+
+    # Ensure input arrays are numpy.ndarrays
```
maboualidev: The comments says, `Ensure input arrays are bumpy.ndarrays`

I am assuming moving passed this point the array is assumed it is `np.ndarrays`.

However, The following code snipper:

```python
if isinstance(lat_aux_grid, xr.DataArray):
         lat_aux_grid = lat_aux_grid.values
```
only checks if the data is `xr.DataArray` it extracts the `numpy` part.

What if the user sends a `list`? Would that cause any trouble, since it is not a `np.ndarray`?

src/geocat/comp/__init__.py
```diff
@@ -354,4 +354,92 @@ def eofunc(data, neval, **kwargs) -> xr.DataArray:
...
+        msg = np.float64(msg)
+    else:
+        msg = np.float32(msg)
+
+
+    # Call ncomp function
+    out_arr = _ncomp._moc_globe_atl(lat_aux_grid, a_wvel, a_bolus, a_submeso,
+                                    tlat, rmlak, msg)
+
+    if meta:
```
maboualidev: `meta` data exists only on `xr.DataArray`. I believe this need to be changed to count that in.

src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -512,3 +513,161 @@ cdef ncomp_attributes_to_dict(libncomp.ncomp_attributes attrs):
...
+    cdef libncomp.ncomp_array* ncomp_a_wvel = np_to_ncomp_array(a_wvel)
+    cdef libncomp.ncomp_array* ncomp_a_bolus = np_to_ncomp_array(a_bolus)
+    cdef libncomp.ncomp_array* ncomp_a_submeso = np_to_ncomp_array(a_submeso)
+    cdef libncomp.ncomp_array* ncomp_tlat = np_to_ncomp_array(tlat)
+    cdef libncomp.ncomp_array* ncomp_rmlak = np_to_ncomp_array(rmlak)
+
+    # Handle missing values
+    missing_inds_a_wvel = None
+
+    if msg is None or np.isnan(msg):    # if no missing value specified, assume NaNs
```
maboualidev: @khallock Does the new `class Array` also handles the missing value assignment to `ncomp_array`? If not I think it would be a good idea to add it to that class.

Currently the workflow is that we get the bulk of the `ncomp_array` from the `np.ndarray` and then assign the missing value to it. It would be nice if we could do both at the same time.

src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -512,3 +513,161 @@ cdef ncomp_attributes_to_dict(libncomp.ncomp_attributes attrs):
         attr = (attrs.attribute_array)[i]
         d[attr.name] = ncomp_to_np_array(attr.value)
     return d
+
+
+
+cdef printNcompArray(libncomp.ncomp_array* ncarr):
```
maboualidev: is `printNcompArray` needed? or is it part of the debugging?

Although, it might be good to have some internal utilities in the library anyway, for internal use. Leave it up to you if you want to keep it or not.

src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -512,3 +513,161 @@ cdef ncomp_attributes_to_dict(libncomp.ncomp_attributes attrs):
...
+    cdef libncomp.ncomp_array* ncomp_output = NULL
+
+    cdef int ier
+    with nogil:
+        ier = libncomp.moc_globe_atl(ncomp_lat_aux_grid, ncomp_a_wvel, ncomp_a_bolus,
+                                  ncomp_a_submeso, ncomp_tlat, ncomp_rmlak,
+                                  &ncomp_output)
+
+    # Check errors ier
+    if ier:
```
maboualidev: Nice. I think I have forgot to check for `ier` in _eofunc. I have to double check that. (no action item here for anyone but me; Seeing this just reminded me I have forgot something)

src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -512,3 +513,161 @@ cdef ncomp_attributes_to_dict(libncomp.ncomp_attributes attrs):
...
+            current_elem_d = data_d[ind]
+            printf("%d ", current_elem_d)
+        else:
+            current_elem_f = data_f[ind]
+            printf("%f ", current_elem_f)
+
+    printf("\n\n")
+
+
+cdef printNumpyArray(np.ndarray arr):
```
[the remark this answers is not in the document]
erogluorhan: printNumpyArray function is aimed at debugging/diagnosis purposes for Cython code. This function is included intentionally within this pull-request, thinking that other contributors might benefit it as well. We can isolate them from the code base though.

src/geocat/comp/__init__.py
erogluorhan: Thanks for this question, as it reminded me missing information in the function description.  The function currently assumes :class:`xarray.DataArray` or :class:`numpy.ndarray` for inputs, inspired by `linint2` function `(xo and yo).` However, it could be a good idea to discuss this as a team whether to assume data types about input arrays or fully handle exceptions.

src/geocat/comp/__init__.py
erogluorhan: To my understanding, `meta` information can be significant only if data type is `xr.DataArray` and output array is somehow related to the primary input array (w.r.t dimensions etc.). However, `moc_globe_atl` original function does not seem to generate such an output array that is so related to any of its input arrays. Therefore, even though inputs are of type `xr.DataArray`, the output still would not have a meta information coming from them. This is why current version of `moc_globe_atl` is written with `meta=False` default argument in contrast to other functions. 

For possible cases where meta can be of interest, the current version of the code has the following TODO item for potential future revisit:

`if meta:
        pass
        # TODO: Retaining possible metadata might be revised in the future`

src/geocat/comp/_ncomp/ncomp.pyx
erogluorhan: printNcompArray function is aimed at debugging/diagnosis purposes for Cython code. This function is included intentionally within this pull-request, thinking that other contributors might benefit it as well. We can isolate them from the code base though.

src/geocat/comp/_ncomp/ncomp.pyx
erogluorhan: Thanks for your comment!

src/geocat/comp/__init__.py
maboualidev: `meta=true` means that if the input data has any meta/attribute associated with it, the same attributes/meta are also attached to the output data. The output data might have some extra attributes or it might not (in the case of moc it does not have any extra). And that's true; the input data could have some meta/attributes if it is of `xr.DataArray`. Currently, `numpy.ndarray` do not have any attributes.

So I think the code needs to be changed to:

```python
if meta and isinstance(input, xr.DataArray):
  # add any data attribute that our already associated to input to the output as well.
```

The `meta` tag is similar to `_wrap` alternative of the functions in the NCL.

If there is no Attribute transfer from input to the output, then we could remove this check then.

src/geocat/comp/_ncomp/ncomp.pyx
maboualidev: Agreed. Let's keep it then. It is good to have these sort of functions for internal use at least.

src/geocat/comp/_ncomp/ncomp.pyx
```diff
@@ -512,3 +513,161 @@ cdef ncomp_attributes_to_dict(libncomp.ncomp_attributes attrs):
...
+
+    cdef int ier
+    with nogil:
+        ier = libncomp.moc_globe_atl(ncomp_lat_aux_grid, ncomp_a_wvel, ncomp_a_bolus,
+                                  ncomp_a_submeso, ncomp_tlat, ncomp_rmlak,
+                                  &ncomp_output)
+
+    # Check errors ier
+    if ier:
+      warnings.warn("moc_globe_atl: {}: There is an error".format(ier), NcompWarning)
```
maboualidev: Why are we generating warning? why not raising error?

src/geocat/comp/_ncomp/ncomp.pyx
khallock: @maboualidev The new `_ncomp.Array` class does not currently handle missing value assignment, but I agree that it would be a good idea for it to do so.

clyne:
@erogluorhan please resolve conflicts in this PR and we'll get it merged. Yay!

clyne:
Need to resolve error/warning discussion raised by @maboualidev

src/geocat/comp/_ncomp/ncomp.pyx
erogluorhan: I agree with this question such that if `ier` is other than 0, then the case is an erroneous one, i.e. the code should be generating an error instead of a warning.

The corresponding correction has been performed by a new commit.

erogluorhan:
@clyne We should be good now to merge this PR because:

1) The error/warning discussion raised by @maboualidev has been resolved by converting warning to error at Cython level.

2) Two utility functions `printNcompArray` and `printNumpyArray` have been removed from Cython source code.

3) Merge conflicts have been resolved.

PR #45, opened 2019-12-02
Adding eofunc_ts

maboualidev:
Adds `eofunc_ts` python function which wraps mcl's `eofunc_ts` and `eofunc_ts_n` and their corresponding `_wrap` functions.

### DIFF
--- a/.circleci/environment-dev-Darwin-3.6.yml
+++ b/.circleci/environment-dev-Darwin-3.6.yml
@@ -10,3 +10,5 @@ dependencies:
   - dask
   - distributed
   - pytest
+  - netcdf4
+  - scipy
--- a/.circleci/environment-dev-Darwin-3.6.yml
+++ b/.circleci/environment-dev-Darwin-3.6.yml
@@ -10,5 +10,4 @@ dependencies:
   - dask
   - distributed
   - pytest
-  - netcdf4
-  - scipy
+
--- a/.circleci/environment-dev-Linux-3.6.yml
+++ b/.circleci/environment-dev-Linux-3.6.yml
@@ -10,5 +10,3 @@ dependencies:
   - dask
   - distributed
   - pytest
-  - netcdf4
-  - scipy
--- a/.circleci/environment-dev-Linux-3.6.yml
+++ b/.circleci/environment-dev-Linux-3.6.yml
@@ -10,3 +10,5 @@ dependencies:
   - dask
   - distributed
   - pytest
+  - netcdf4
+  - scipy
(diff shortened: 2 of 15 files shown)


.circleci/environment-dev-Darwin-3.6.yml
```diff
@@ -10,3 +10,5 @@ dependencies:
   - dask
   - distributed
   - pytest
+  - netcdf4
+  - scipy
```
khallock: Is there a reason `netcdf4` and `scipy` have been added as dependencies? I can't find any code that actually uses either of them.

.circleci/environment-dev-Linux-3.6.yml
```diff
@@ -10,3 +10,5 @@ dependencies:
   - dask
   - distributed
   - pytest
+  - netcdf4
+  - scipy
```
khallock: See previous comment.

.circleci/run-checks.sh
```diff
@@ -4,4 +4,7 @@ set -e
 set -eo pipefail
 
 source activate ${ENV_NAME}
+pip install netcdf4
+pip install scipy
```
khallock: See previous comment.

Also note that these packages will already have been installed via conda at this point because they're in the environment-dev-*.yml files.

docs/conda_environment.yml
```diff
@@ -8,3 +8,5 @@ dependencies:
   - xarray
   - dask
   - ncomp
+  - scipy
+  - netcdf4
```
khallock: See previous comment.

.circleci/environment-dev-Darwin-3.6.yml
maboualidev: Actually that doesn't work. Somehow they don't get installed.

I am removing them.

.circleci/environment-dev-Linux-3.6.yml
maboualidev: It's removed

.circleci/run-checks.sh
maboualidev: These are needed for the test only.

The sample data is in NetCDF Format and `xr.open_dataset` apparently requires them.

docs/conda_environment.yml
maboualidev: Removed them from the yml

maboualidev:
@khallock @clyne (1) Added the examples to the doc-string and documentation and (2) Changed the dependencies for tests.

PR #46, opened 2019-12-03
Update moc_globe_atl to use _ncomp.Array class

khallock:

PR #47, opened 2019-12-03
Reorganize Cython code and add unit testing

khallock:
There are now two .pxd files included with the Cython code:
```
src/geocat/comp/libncomp.pxd
src/geocat/comp/_ncomp.pxd
```

`libncomp.pxd` should be used for declaring NComp C-API function signatures to allow the `_ncomp.pyx` module to `cimport libncomp` and then call the NComp functions using the `libncomp` namespace.

As of this commit, `_ncomp.pxd` is only being used to provide a class definition for the recently-added `_ncomp.Array` class (so that other Cython modules/programs can `cimport _ncomp` and then access the class as `_ncomp.Array`), but could be extended to include other "cdef"ed functionality from `_ncomp.pyx` in the future. Note that "cdef"ed functions or variables in Cython are only accessible from Cython and are not directly visible within Python.

A new unit test that compiles and tests Cython code built against the internal `geocat.comp._ncomp` module has been added: `test/test_cython.py`
This test automatically compiles and runs the tests listed in the file `test/cython_tests/test_cython.pyx`.


khallock:
Note that merging #46 should resolve the merge conflict for this PR if it's approved first.

clyne:
@khallock your PR has conflicts now.

khallock:
@clyne Thanks for the heads up, the merge conflicts should be resolved now.

PR #48, opened 2019-12-10
ndpolyfit

maboualidev:
`geocat-comp.ndpolyfit` extends `numpy.polyfit` functionality. The main differences are:

- `np.polyfit` accepts a one dimensional array or two-dimensional array, `geocat-comp.ndpolyfit` accepts any dimensional array.

- `np.polyfit` does not handle the missing values and if there are any `NaN` in the data, everything turns into `NaN`, while `geocat-comp.ndpolyfit` recognizes missing values.

- in addition to accepting `array-like` inputs, `geocat-comp.ndpolyfit` make use of `Dask` array and if they are chunked.

This merge adds at least 24 extra unit tests.


setup.py
```diff
@@ -33,5 +33,5 @@
     package_dir={'': 'src'},
     namespace_packages=['geocat'],
     packages=["geocat.comp"],
-    version=__version__,
+    version=__version__, install_requires=['numpy', 'xarray', 'dask[complete]']
```
khallock: For consistency, `install_requires=['numpy', 'xarray', 'dask[complete]']` should be on its own line.

maboualidev: reformat is done. It is now on a separate line.
It was added automatically by PyCharm.

clyne:
@maboualidev FYI your PR now has conflicts.

maboualidev:
> @maboualidev FYI your PR now has conflicts.

Thank you @clyne ; I resolved the conflict now.

PR #49, opened 2019-12-14
Ndpolyval

maboualidev:
Extends `numpy.polyval` in support of `geocat.comp.ndpolyfit` which supports multi-dimensional arrays (including Xarray) and Dask support.


src/geocat/comp/polynomial.py
```diff
@@ -1,10 +1,69 @@
...
+            missing_value = data.attrs["_FillValue"]
+        else:
+            missing_value = np.nan
+    else:
+        missing_value = np.nan
+
+    return missing_value
+
+
+def _unchuck_ifneeded(data: da.Array, axis: int) -> da.Array:
```
clyne: Should "_unchuck" be "_unchunck"?

src/geocat/comp/polynomial.py
```diff
@@ -1,10 +1,69 @@
-from typing import Iterable
+from typing import Iterable, Any
 
 import numpy as np
 import xarray as xr
 import dask.array as da
 
 
+def _get_missing_value(data: xr.DataArray, args: dict) -> Any:
```
clyne: Is the use of this function consistent with how we handle missing values with GeoCAT functions that wrap compiled (Fortran) code? Just checking.

src/geocat/comp/polynomial.py
```diff
@@ -334,6 +379,116 @@ def _reverse_rearrange_axis(data: np.ndarray, axis, trailing_shape: tuple) -> np
...
+    return data
+
+
+def ndpolyval(p: Iterable, x: Iterable = None, axis: int = 0, **kwargs):
+    """
+    Extended version of `numpy.polyval` to support multi-dimensional outputs provided by `geocat.comp.ndpolyfit`.
+
+    As the name suggest, this version supports a multi-dimensional `p` array. Let's say `p` is of dimension `(s0,s1,s2)`
+    and `axis=1`, then the output would be of dimension `(s0, M, s2)` where M depends on `x`.
+    The same way, you `x` could be a multi dimensional array or a single array. In another word, `x` is either of
```
clyne: Typo: "you" -> "your"? Or perhaps just delete "you"

src/geocat/comp/polynomial.py
```diff
@@ -334,6 +379,116 @@ def _reverse_rearrange_axis(data: np.ndarray, axis, trailing_shape: tuple) -> np
...
+    dimension `(M, )`, `(M, 1)`, `(1, M)` or, in our example, of dimension `(s0, M, s2)`. When `x` is not the vector,
+    is must have the same dimension as of `p` except for the dimension that is defined by `axis`.
```
clyne: Typo: "is" -> "it"

src/geocat/comp/polynomial.py
```diff
@@ -334,6 +379,116 @@ def _reverse_rearrange_axis(data: np.ndarray, axis, trailing_shape: tuple) -> np
...
+        x (:class:`Iterable`):
+            the coordinates where polynomial must be evaluated
+
+        axis (:class:`int`):
+        The axis where the polynomials are there.
+
+        **kwargs:
+            Currently not used.
+
+    Returns (:class: `xr.DataArray`:
```
clyne: How about a short usage example?

src/geocat/comp/polynomial.py
maboualidev: True; 😄 . Fixed it.

src/geocat/comp/polynomial.py
maboualidev: Short answer: Yes It is consistent.

Now the long answer: 😄 
first a side note: These functions don't end up calling any FORTRAN code. 

This is how the missing_value is retrieved:

- Checking if the user has manually provided any information on it. Use that if the use has provided it.
- If the user has not provided any information; but the array is of type `Xarray` which it could have attributes, we are going to check if there is a property called `missing_value` or `_FillValue`. If there is a value provided for these attributes, that is used as missing_value. If both are provided `missing_value` get's precedence. This is the same as in NCL which is based on CF-convention.
- If the user hasn't manually provided any value for the missing_value, or the array is not `xarray`, i.e. doesn't have attributes, eg. if it is `np.ndarray` it cannot have any attributes, or if it is `Xarray` but does not have any of the proper attributes for missing value available, we are assuming that `np.nan` is the default missing value.

Not applicable in this code because there is no FORTRAN code for these functions: but prior to sending `np.nan` to the FORTRAN code, at the python level, we do have a function that we will convert the `np.nan` to a proper default numeric missing value. The function is called `get_dafault_fill`. Currently we make use of this to convert `np.nan` to something that the FORTRAN level code would recognize as missing value.

src/geocat/comp/polynomial.py
maboualidev: removed `you`.

src/geocat/comp/polynomial.py
maboualidev: changed `is` to `it`

src/geocat/comp/polynomial.py
maboualidev: added example usage with some explanation as comment.

src/geocat/comp/polynomial.py
erogluorhan: Indeed, should be "_unchunk" instead of "_unchunck" 😬 

(The implementation and test cases looks food to me and already approved though)

PR #50, opened 2019-12-20
rcm2points

abb58:
Infrastructure for rcm2points

### DIFF
--- a/src/geocat/comp/__init__.py
+++ b/src/geocat/comp/__init__.py
@@ -581,3 +581,139 @@ def moc_globe_atl(lat_aux_grid, a_wvel, a_bolus, a_submeso, tlat, rmlak,
         out_arr = xr.DataArray(out_arr)
 
     return out_arr
+
+
+def rcm2points(lat2d, lon2d, fi, lat1d, lon1d, msg=None, meta=False):
+     """Interpolates data on a curvilinear grid (i.e. RCM, WRF, NARR) to a rectilinear grid.
+
+     Args:
+
+         lat2d (:class:`numpy.ndarray`):
+ 	    A two-dimensional array that specifies the latitudes locations
+ 	    of fi. Because this array is two-dimensional it is not an associated
+ 	    coordinate variable of `fi`. The latitude order must be south-to-north.
+
+         lon2d (:class:`numpy.ndarray`):
+ 	    A two-dimensional array that specifies the longitude locations
+ 	    of fi. Because this array is two-dimensional it is not an associated
+ 	    coordinate variable of `fi`. The latitude order must be west-to-east.
+
+         fi (:class:`numpy.ndarray`):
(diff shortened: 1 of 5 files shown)


src/geocat/comp/__init__.py
```diff
@@ -581,3 +585,127 @@ def moc_globe_atl(lat_aux_grid, a_wvel, a_bolus, a_submeso, tlat, rmlak,
...
+
+    # ensure lat1d and lon1d are numpy.ndarrays
+    if isinstance(lat1dPoints, xr.DataArray):
+        lat1dPoints = lat1dPoints.values
+    if isinstance(lon1dPoints, xr.DataArray):
+        lon1dPoints = lon1dPoints.values
+
+    fi_data = fi.data
+
+    if isinstance(fi_data, np.ndarray):
```
khallock: Is there any reason why `rcm2points` can only be used with NumPy arrays but not Dask arrays?

abb58: Addressed. https://github.com/NCAR/geocat-comp/pull/50/commits/3eb71dc31c2b52849e0acdce0074622b7e0cbe0a

src/geocat/comp/__init__.py
```diff
@@ -581,3 +585,127 @@ def moc_globe_atl(lat_aux_grid, a_wvel, a_bolus, a_submeso, tlat, rmlak,
...
+	    #	       an other .nc file.
+	    newlat1D_points=np.linspace(lat2D_curv.min(), lat2D_curv.max(), 100)
+	    newlon1D_points=np.linspace(lon2D_curv.min(), lon2D_curv.max(), 100)
+
+	    ht_points = geocat.comp.rcm2points(lat2D_curv, lon2D_curv, ht_curv, newlat1D_points, newlon1D_points)
+
+    """
+
+    # Basic sanity checks
+    if lat2d.shape[0] != lon2d.shape[0] or lat2d.shape[1] != lon2d.shape[1]:
```
erogluorhan: Code and tests are LGTM except a little point for these basic sanity check lines:

If lat2d and/or lon2d comes as 1D or None, these lines will throw your `DimensionError ` saying that  

> The input lat/lon grids must be the same size

But the actual error will be that shape[1] won't be accessed. My suggestion is that you can either

1) Put a TODO comment

or

2) Create another Exception other than DimensionError (you can even fill that only with `pass` line)

for future reference.

erogluorhan: I noticed afterwards that I wrongly assumed your `if condition` as a `try` block. Current version would not have the problem in my original comment such that if arguments come as None or 1D, builtin IndexError would be thrown.

clyne:
@abb58 this PR is approved pending resolution of conflicts.

abb58:
> @abb58 this PR is approved pending resolution of conflicts.

@clyne All the conflicts are addressed but everyone's approval is dismissed.

PR #53, opened 2019-12-27
triple2grid

abb58:
triple2grid and grid2triple

### DIFF
--- a/src/geocat/comp/__init__.py
+++ b/src/geocat/comp/__init__.py
@@ -1265,3 +1265,88 @@ def triple2grid(x, y, data, xgrid, ygrid, **kwargs):
         fo = xr.DataArray(fo)
 
     return fo
+
+
+def grid2triple(x, y, z, msg=None, meta=False):
+    """Converts a two-dimensional grid with one-dimensional coordinate variables
+       to an array where each grid value is associated with its coordinates.
+
+    Args:
+
+	x (:class:`numpy.ndarray`):
+            Coordinates associated with the right dimension of the variable `z`.
+            It must be the same dimension size (call it mx) as the right
+            dimension of `z`.
+
+	y (:class:`numpy.ndarray`):
+            Coordinates associated with the left dimension of the variable `z`.
+            It must be the same dimension size (call it ny) as the left
+            dimension of `z`.
+
+	z (:class:`numpy.ndarray`):
+            Two-dimensional array of size ny x mx containing the data values.
+            Missing values may be present in `z`, but they are ignored.
+
+	msg (:obj:`numpy.number`):
+	    A numpy scalar value that represent a missing value in `z`.
+	    This argument allows a user to use a missing value scheme
+	    other than NaN or masked arrays, similar to what NCL allows.
+
+	meta (:obj:`bool`):
+	    Set to False to disable metadata; default is False.
+
+    Returns:
+	:class:`numpy.ndarray`: If any argument is "double" the return type
+        will be "double"; otherwise a "float" is returned.
+
+    Description:
+        The maximum size of the returned array will be 3 x ld where ld <= ny*mx.
+        If no missing values are encountered in z, then ld=ny*mx. If missing
+        values are encountered in z, they are not returned and hence ld will be
+        equal to ny*mx minus the number of missing values found in z. The return
+        array will be double if any of the input arrays are double, and float
+        otherwise.
+
+    Examples:
+
+	Example 1: Using grid2triple with :class:`xarray.DataArray` input
+
+	.. code-block:: python
+
+	    import numpy as np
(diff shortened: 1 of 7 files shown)


src/geocat/comp/__init__.py
```diff
@@ -1105,4 +1105,248 @@ def rcm2points(lat2d, lon2d, fi, lat1dPoints, lon1dPoints, opt=0, msg=None, meta
...
+    if isinstance(ygrid, xr.DataArray):
+        ygrid = ygrid.values
+
+    if isinstance(data, np.ndarray):
+        fo = _ncomp._triple2grid(x, y, data, xgrid, ygrid, options, msg)
+    else:
+        raise TypeError
+
+    if meta and isinstance(input, xr.DataArray):
+        pass	 # TODO: Retaining possible metadata might be revised in the future
```
clyne: Rather than silently ignoring this option it would probably be better to throw an error, or remove it. What do others think?

src/geocat/comp/__init__.py
```diff
@@ -1105,4 +1105,248 @@ def rcm2points(lat2d, lon2d, fi, lat1dPoints, lon1dPoints, opt=0, msg=None, meta
...
+    if isinstance(z, xr.DataArray):
+        z = z.values
+
+    if isinstance(z, np.ndarray):
+        fo = _ncomp._grid2triple(x, y, z, msg)
+    else:
+        raise TypeError
+
+    if meta and isinstance(input, xr.DataArray):
+        pass	 # TODO: Retaining possible metadata might be revised in the future
```
clyne: Same comment here wrt silently ignoring this option.

src/geocat/comp/__init__.py
abb58: It would be logical to throw an exception to have an informed feedback that NCL `_wrap()` functions are not yet supported.

abb58: https://github.com/NCAR/geocat-comp/pull/53/commits/0d8d4f53013fd903ee03df52ace649a58ea354fa

src/geocat/comp/__init__.py
abb58: https://github.com/NCAR/geocat-comp/pull/53/commits/0d8d4f53013fd903ee03df52ace649a58ea354fa

src/geocat/comp/__init__.py
```diff
@@ -1105,4 +1110,248 @@ def rcm2points(lat2d, lon2d, fi, lat1dPoints, lon1dPoints, opt=0, msg=None, meta
...
+        data = data.values
+    if isinstance(xgrid, xr.DataArray):
+        xgrid = xgrid.values
+    if isinstance(ygrid, xr.DataArray):
+        ygrid = ygrid.values
+
+    if isinstance(data, np.ndarray):
+        fo = _ncomp._triple2grid(x, y, data, xgrid, ygrid, options, msg)
+    else:
+        raise TypeError
```
khallock: We should probably add an actual error message in addition to throwing an exception, maybe something like:
```
raise TypeError("ERROR: triple2grid: unsupported input type {} for 'data' argument.".format(type(data)))

File: src/geocat/comp/__init__.py [hunk 4]
```diff
@@ -1105,4 +1110,248 @@ def rcm2points(lat2d, lon2d, fi, lat1dPoints, lon1dPoints, opt=0, msg=None, meta
...
+        x = x.values
+    if isinstance(y, xr.DataArray):
+        y = y.values
+    if isinstance(z, xr.DataArray):
+        z = z.values
+
+    if isinstance(z, np.ndarray):
+        fo = _ncomp._grid2triple(x, y, z, msg)
+    else:
+        raise TypeError
```
khallock: Same as above, we should consider adding an error message to go with the exception.

test/test_dpres_plevel.py
```diff
@@ -25,8 +25,3 @@ def test_rcm2points_float64(self):
         result_dp = geocat.comp.dpres_plevel(plev, psfc)
         np.testing.assert_array_equal(expected_dp, result_dp.values)
 
-class Test_dpres_plevel_float32(ut.TestCase):
-    def test_dpres_plevel_float32(self):
-        plev_asfloat32 = plev.astype(np.float32)
-        result_dp = geocat.comp.dpres_plevel(plev_asfloat32, psfc)
-        np.testing.assert_array_equal(expected_dp, result_dp.values)
```
khallock: Was this test removed intentionally? This test is for `dpres_plevel` and appears to be unrelated to the `triple2grid` PR.

khallock: I'm going to create a new PR that re-enables this test so we can diagnose the issue with `dpres_plevel` without holding up this PR.

PR #54, opened 2020-01-22
Dtrend + More

maboualidev:
Adds dtrend
+ Document/examples for ndpolyfit/ndpolyval/dtrend.

### DIFF
--- a/src/geocat/comp/polynomial.py
+++ b/src/geocat/comp/polynomial.py
@@ -508,7 +508,11 @@ def detrend(data: Iterable, deg=1, axis=0, **kwargs):
 
     x = kwargs.get("x", None)
     if x is None:
-        x = np.arange(data_shape[axis], dtype=float)
+        x = np.arange(data_shape[axis], dtype=np.float)
+    elif not isinstance(x, np.ndarray):
+        x = np.asarray(x).astype(np.float)
+
+    x = x.reshape((-1, ))
 
     p = ndpolyfit(x, data, deg, axis, missing_value=missing_value)
 
--- a/src/geocat/comp/polynomial.py
+++ b/src/geocat/comp/polynomial.py
@@ -500,6 +500,37 @@ def ndpolyval(p: Iterable, x: Iterable, axis: int = 0, **kwargs):
     Returns (:class: `xr.DataArray`:
         polynomial evaluated with the provided coordinates.
 
+    Examples:
+
+        * Evaluating a polynomial:
+
+        >>> p = [2, 3] # representing y = 2*x + 3
+        >>> x = np.arange(-5, 6, dtype="float")
+        >>> x
+        array([-5., -4., -3., -2., -1.,  0.,  1.,  2.,  3.,  4.,  5.])
+        >>> from geocat.comp.polynomial import ndpolyval
+        >>> y = ndpolyval(p, x)
+        >>> y.shape
+        (11,)
+        >>> print(y)
+        <xarray.DataArray (dim_0: 11)>
+        array([-7., -5., -3., -1.,  1.,  3.,  5.,  7.,  9., 11., 13.])
+        Dimensions without coordinates: dim_0
+        >>> np.testing.assert_almost_equal(y, 2*x+3)
+
+        * evaluating a multi-dimensional fitted polynomial:
+
+        >>> p = np.tile(np.asarray(p).reshape(1, 2, 1, 1), [3, 1, 4, 5])
+        >>> p.shape
+        (3, 2, 4, 5)
+        >>> p[1, :, 1, 1]
+        array([2, 3])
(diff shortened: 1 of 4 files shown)


src/geocat/comp/polynomial.py
```diff
@@ -113,6 +112,81 @@ def ndpolyfit(x: Iterable, y: Iterable, deg: int, axis: int = 0, **kwargs) -> (x
...
+        Attributes:
+            deg:             2
+            provided_rcond:  None
+            full:            False
+            weights:         None
+            covariance:      False
+
+        * Fitting polynomial with missing values:
+
+        >>> # Let's introduce some missing values:
```
clyne: ndpolyfit will recognized NaNs as missing values, correct? Perhaps clarify this comment such as:

`# Ordinarily NaN's are treated as missing values. In this example let's introduce a different value to indicate missing data.`

src/geocat/comp/polynomial.py
```diff
@@ -545,6 +649,95 @@ def _ndpolyval(p: np.ndarray, x: np.ndarray, axis: int = 0, **kwargs) -> np.ndar
     return y
 
 
+def detrend(data: Iterable, deg=1, axis=0, **kwargs):
+    """
+    This method, at the minimum, provides all the functionality that is provided by NCL's 'dtrend',
```
clyne: For users not familiar with NCL it would be good to introduce this documentation with more general statement about what the function does. For example, 

`# Estimates and removes the least squares linear, quadratic, or higher order trend from an array....`

src/geocat/comp/polynomial.py
```diff
@@ -545,6 +649,95 @@ def _ndpolyval(p: np.ndarray, x: np.ndarray, axis: int = 0, **kwargs) -> np.ndar
...
+    'dtrend_quadratic', 'dtrend_quadratic_msg_n', 'dtrend_msg_n', 'dtrend_msg', 'dtrend_n'.
+    However, this function is not limited to quadratic detrending and you could use higher polynomial degree as well.
+
+    Args:
+        data (:class:`array_like`):
+            a multi-dimensional numeric array
+
+        deg (:class:`int`, optional):
+            a non-negative integer determining the degree of the polynomial to use for detrending. Degault value is 1.
```
clyne: Typo: degault -> default

src/geocat/comp/polynomial.py
```diff
@@ -545,6 +649,95 @@ def _ndpolyval(p: np.ndarray, x: np.ndarray, axis: int = 0, **kwargs) -> np.ndar
...
+            missing_value (:class:`numeric`)
+                A value that must be ignored. Default is NaN.
+
+    Returns:
+          an `xarray.DataArray` containing the detrended data.
+
+    Examples:
+
+        * Detrending a data:
+
```
clyne: What package import statements are needed for users to call ndpolyfit() and dtrend()?

src/geocat/comp/polynomial.py
maboualidev: done

src/geocat/comp/polynomial.py
maboualidev: done

erogluorhan:
Is it intentional to define function name as "detrend" instead of its original NCL name, "dtrend"?

src/geocat/comp/polynomial.py
maboualidev: Added the packages in the example

src/geocat/comp/polynomial.py
maboualidev: done

maboualidev:
> Is it intentional to define function name as "detrend" instead of its original NCL name, "dtrend"?

It is intentional. But I am open to change it to "dtrend". The reason I chose to call it "detrend" is that this is actual English verb. while "dtrend" is not.

But I am ok either way.

erogluorhan:
It's a tradeoff between keeping NCL function names as is for easy tracking and your suggestion (which also makes sense to me). It is up to you to discuss this with other team members, though I am approving the PR now because the implementation and testing look good to me.

maboualidev:
> It's a tradeoff between keeping NCL function names as is for easy tracking and your suggestion (which also makes sense to me). It is up to you to discuss this with other team members, though I am approving the PR now because the implementation and testing look good to me.

`detrend` is providing the functionality for all these NCL methods:

- 'dtrend', 
- 'dtrend_quadratic', 
- 'dtrend_quadratic_msg_n', 
- 'dtrend_msg_n', 
- 'dtrend_msg', 
- 'dtrend_n'

plus it does some more. e.g. you are not limited to only degree 0, 1, and 2 detrending. All these functions are collapsed into one single python function. 
However, I am completely agnostic about the use of `dtrend` or `detrend`. I am ok both ways.

PR #58, opened 2020-03-09
Linint2points

erogluorhan:
This PR adds `linint2_points` functionality to geocat-comp level as well as its comprehensive test cases (Including output comparisons to NCL but excluding parallelization) with `test_linint2points.py`.

This PR should be good now to be built since its corresponding **ncomp** PR ([https://github.com/NCAR/ncomp/pull/24](url))is merged as well as the erroneous `test_dpres_plevel.py` has been removed from develop branch.


src/geocat/comp/__init__.py
```diff
@@ -1142,13 +1141,204 @@ def rcm2points(lat2d, lon2d, fi, lat1dPoints, lon1dPoints, opt=0, msg=None, meta
...
 
+def linint2_points(fi, xo, yo, icycx, msg=None, meta=False, xi=None, yi=None):
+    """Interpolates from a rectilinear grid to an unstructured grid or locations using bilinear interpolation.
+
+    Args:
+
+        fi (:class:`xarray.DataArray` or :class:`numpy.ndarray`):
+            An array of two or more dimensions. The two rightmost
+            dimensions (nyi x nxi) are the dimensions to be used in
+            the interpolation. If missing values are present, the
```
clyne: Is this statement accurate:

"If missing values are present, the value of `msg` must be set appropriately."

If the arrays use NaNs as missing values does 'msg' require setting?

src/geocat/comp/_ncomp.pyx
```diff
@@ -1134,6 +1134,110 @@ def _rgrid2rcm(np.ndarray lat1d_np, np.ndarray lon1d_np, np.ndarray fi_np, np.nd
...
+
+    Interpolates from a rectilinear grid to an unstructured grid
+    or locations using bilinear interpolation.
+
+    Args:
+
+        fi (:class:`xarray.DataArray` or :class:`numpy.ndarray`):
+            An array of two or more dimensions. The two rightmost
+            dimensions (nyi x nxi) are the dimensions to be used in
+            the interpolation. If missing values are present, the
```
clyne: Same comment as previously wrt setting of 'msg' argument

src/geocat/comp/__init__.py
erogluorhan: This statement is taken from the argument description in the NCL counterpart:

[https://www.ncl.ucar.edu/Document/Functions/Built-in/linint2_points.shtml](url)

I think that the user-defined missing values other than NaNs are meant in this statement.

We can replace it with a clearer one, though. How about the following:

"If user-defined missing values are present (other than NaNs), the value of `msg` must be set appropriately."

src/geocat/comp/_ncomp.pyx
erogluorhan: This statement is taken from the argument description in the NCL counterpart:

[https://www.ncl.ucar.edu/Document/Functions/Built-in/linint2_points.shtml](url)

I think that the user-defined missing values other than NaNs are meant in this statement.

We can replace it with a clearer one, though. How about the following:

"If user-defined missing values are present (other than NaNs), the value of `msg` must be set appropriately."

src/geocat/comp/__init__.py
clyne: My understanding is that the  behavior of the GeoCAT functions is slightly different than the NCL versions because the Python variants recognize NaNs as missing values by default, whereas with NCL you have to explicitly identify the missing value identifier. 

In any event I like you suggested change. Please proceed. Thanks.

src/geocat/comp/_ncomp.pyx
clyne: Looks good.

src/geocat/comp/__init__.py
erogluorhan: Thanks for the clarification; I agree with it. Pushed the addressing commit for it.

src/geocat/comp/_ncomp.pyx
erogluorhan: Thanks for the clarification. Pushed the addressing commit for it.

src/geocat/comp/__init__.py
```diff
@@ -1142,13 +1141,205 @@ def rcm2points(lat2d, lon2d, fi, lat1dPoints, lon1dPoints, opt=0, msg=None, meta
...
+        yi = yi.values
+
+    if isinstance(xo, xr.DataArray):
+        xo = xo.values
+    if isinstance(yo, xr.DataArray):
+        yo = yo.values
+
+    fi_data = fi.data
+
+    if isinstance(fi_data, np.ndarray):
```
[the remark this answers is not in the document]
erogluorhan: Great catch, thanks for this! Addressed this and tested in the newly implemented test cases.

test/test_linint2points.py
```diff
@@ -0,0 +1,216 @@
...
+    #     fi = xr.DataArray(self._fi_np[:, :, ::-1, :], dims=['time', 'level', 'lat', 'lon'], coords={'lat': self._yi_reverse, 'lon': self._xi}).chunk(_chunks)
+    #     with self.assertWarns(geocat.comp._ncomp.NcompWarning):
+    #         geocat.comp.linint2_points(fi, self._xo, self._yo, 0).compute()
+
+    def test_linint2points_non_monotonic_np(self):
+        with self.assertWarns(geocat.comp._ncomp.NcompWarning):
+            geocat.comp.linint2_points(self._fi_np[:, :, ::-1, :], self._xo, self._yo, 0, xi=self._xi, yi=self._yi_reverse)
+
+
+# class Test_linint2points_float64(ut.TestCase, BaseTestClass):
```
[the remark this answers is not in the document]
erogluorhan: Another great catch! Added the suggested test cases and verified that function works without explicitly passing `xi` and `yi` as arguments when `fi` is an `xarray.DataArray`.  Also used chunked `xarray.DataArray` as input to verify `isinstance` check works.

@khallock  In addition to suggested test cases in this comment, I have also added two other test cases where inputs with NaN and user-defined (-99) missing values were tested. When implementing these cases, I have found an incident in the code, which might also indicate erroneous outputs for similar inputs for the other functions such as `linint2`, `_grid2triple`, `_triple2grid`, `_rgrid2rcm`, `_rcm2rgrid`, `_rcm2points`, `_dpres_plevel`, and probably others. I had a chance to test `linint2` for the same issue and it seems to have the same issue, too. 

**Incident**: When a user-defined missing value (say -99) is present in the input, the output missing values were being converted to NaN (instead of -99) in our Python implementations; however, NCL uses the same missing value of input in outputs (i.e. -99. This was due to the following line, which exists near the end of almost all of our functions at **Cython** level (this exact line is shown from `linint2points` but it is pretty much the same or similar for other functions):

`fo.numpy[fo.numpy == fo_msg] = np.nan
`
Please see my correction in `linint2points` within my very recent commit to see how I have overcome it. Test cases of linint2points generate exactly the same output with NCL with respect to missing values now.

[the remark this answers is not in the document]
clyne: Good catch @erogluorhan . Can you open an issue to get this fixed in the rest of our functions? Thanks!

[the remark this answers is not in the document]
erogluorhan: Thanks! I have opened the issue [https://github.com/NCAR/geocat-comp/issues/60](url)

clyne:
Waiting on examples before merging

erogluorhan:
@khallock and @clyne : This PR is now ready for final reviews with the newly added matplotlib example script of it ([mpl_linint2points.py](https://github.com/NCAR/geocat-comp/blob/linint2points/docs/_static/scripts/examples/matplotlib/mpl_linint2points.py)).

In addition, while addressing our user-defined missing value solution to linint2points that we discussed in the weekly team meeting of 3/31/2020, I have observed a bug for functions `rcm2points, rcm2rgrid, rgrid2rcm` that you can see in detail at [Issue #60](https://github.com/NCAR/geocat-comp/issues/60). If it make sense to you, I can resolve it by a separate branch and PR.

docs/_static/scripts/examples/matplotlib/mpl_linint2points.py
```diff
@@ -0,0 +1,96 @@
...
+import xarray as xr
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+from matplotlib import cm
+from cartopy.mpl.ticker import LongitudeFormatter, LatitudeFormatter
+from cartopy.mpl.geoaxes import GeoAxes
+from mpl_toolkits.axes_grid1 import AxesGrid
+
+# Open a netCDF data file (Sea surface temperature) using xarray default engine and load the data into xarrays
+ds = xr.open_dataset("sst.nc")
```
clyne, line 25: Should we be using Kevin's awesome data fetching utility rather than a local file path?

docs/_static/scripts/examples/matplotlib/mpl_linint2points.py
```diff
@@ -0,0 +1,96 @@
...
+# Call `linint2_points` from `geocat-comp`
+newsst = geocat.comp.linint2_points(sst, newlon, newlat, False)
+
+# Generate figure and set its size (width, height) in inches
+fig = plt.figure(figsize=(10,8))
+
+# Generate Axes grid using a Cartopy projection
+projection = ccrs.PlateCarree()
+axes_class = (GeoAxes, dict(map_projection=projection))
+axgr = AxesGrid(fig, 111, axes_class=axes_class,
```
clyne, line 45: Is there a reason for not using the geocat-viz convenience functions to facilitate plot set up? Just asking.

src/geocat/comp/__init__.py
```diff
@@ -417,8 +417,7 @@ def rcm2rgrid(lat2d, lon2d, fi, lat1d, lon1d, msg=None, meta=False):
                         " a dask.array.Array.")
 
     if meta and isinstance(input, xr.DataArray):
-        pass
-        # TODO: Retaining possible metadata might be revised in the future
+        raise MetaError("ERROR rcm2rgrid: retention of metadata is not yet supported !")
```
clyne, line 420: Can you briefly summarize what the difficulty is in supporting the 'meta' option? Is it something that we can realistically ever support? Should we consider removing the option entirely, or supporting it in a limited sense? Just trying to understand the issue.

src/geocat/comp/__init__.py
```diff
@@ -1142,13 +1141,206 @@ def rcm2points(lat2d, lon2d, fi, lat1dPoints, lon1dPoints, opt=0, msg=None, meta
...
+            the way around the globe. For example, if your longitude
+            values go from, say, -179.75 to 179.75, or 0.5 to 359.5,
+            then you would set this to True.
+
+        msg (:obj:`numpy.number`):
+            A numpy scalar value that represent a missing value in fi.
+            This argument allows a user to use a missing value scheme
+            other than NaN or masked arrays, similar to what NCL allows.
+
+        meta (:obj:`bool`):
```
clyne, line 1185: Should the documentation state that 'meta' isn't supported?

docs/_static/scripts/examples/matplotlib/mpl_linint2points.py
erogluorhan: Great question, John! Indeed, I did have the same question in my mind but did not want to add those dependencies (i.e. `geocat-datafiles`, and also `geocat-viz` as asked in your below comment) to `geocat-comp` on my own. Hi @khallock , it appears to be the correct time to discuss it now: What do you think if we should make use of `geocat-datafiles` and `geocat-viz` in our `geocat-comp` example scripts?

Additionally, I did forget to upload an output plot of the example script FYI, which should have been as follows:
mpl_linint2points

docs/_static/scripts/examples/matplotlib/mpl_linint2points.py
erogluorhan: ibid

src/geocat/comp/__init__.py
erogluorhan: Yes, I agree with this. At least for the current version, until `meta` is resolved for many functions, I have added the following `docstring` to those functions without `meta` handling:

```
meta (:obj:`bool`):
            Set to False to disable metadata since it is not supported
            for the current version; default is True. It will be resolved
            in a future version.
```

as well as I have added a `todo` comment to the same functions to revisit for this issue.

clyne:
> @khallock and @clyne : This PR is now ready for final reviews with the newly added matplotlib example script of it ([mpl_linint2points.py](https://github.com/NCAR/geocat-comp/blob/linint2points/docs/_static/scripts/examples/matplotlib/mpl_linint2points.py)).
> 
> In addition, while addressing our user-defined missing value solution to linint2points that we discussed in the weekly team meeting of 3/31/2020, I have observed a bug for functions `rcm2points, rcm2rgrid, rgrid2rcm` that you can see in detail at [Issue #60](https://github.com/NCAR/geocat-comp/issues/60). If it make sense to you, I can resolve it by a separate branch and PR.

I agree this should be resolved by a separate issue/PR

src/geocat/comp/__init__.py
clyne: Let's discuss the 'meta' issue at our Tuesday meeting.

Regarding the docstring: Sorry, I still find the description confusing.  I've scanned through some of the other functions and it looks like we are, unfortunately, both inconsistent and not terribly coherent in the description of what this thing does.  We should strive to use the same language for all of the functions, and my suggestion would be something like:

meta (:obj:`bool`):
            If set to True and the input array is an Xarray the metadata from the input array will be copied to the output array; default is False. 

In the case of linit2points we can add to the doc string something like: "Warning: this option is not currently supported"

Thoughts?

erogluorhan: Thanks for the docstring suggestions; they sound clearer to me, too. It looks like we should go through docstrings of all the functions and refactor their language and content as much as possible for standardization. This can be performed when we are handling [Issue #60](https://github.com/NCAR/geocat-comp/issues/60)

src/geocat/comp/__init__.py
clyne: sounds good.

src/geocat/comp/__init__.py
clyne: Agreed.

docs/_static/scripts/examples/matplotlib/mpl_linint2points.py
khallock: I agree, we should definitely make use of `geocat-datafiles` and `geocat-viz` here.

Also, I think we may want to consider adding these scripts to the [GeoCAT-examples gallery](https://geocat-examples.readthedocs.io/en/latest/), either in addition to this or as an alternative.

erogluorhan: The [most recent commit](https://github.com/NCAR/geocat-comp/commit/8594cb77f4780c5cd68822f2897c9915b92ffeb4) introduces the use of `geocat-datafiles` and `geocat-viz` in our `mpl_linint2points.py` example script. This PR should be good now for approval and merge (unless there is anything else)

Final version of the plot looks like:

Figure_5

docs/_static/scripts/examples/matplotlib/mpl_linint2points.py
erogluorhan: The [most recent commit](https://github.com/NCAR/geocat-comp/commit/8594cb77f4780c5cd68822f2897c9915b92ffeb4) introduces the use of `geocat-datafiles` and `geocat-viz` in our `mpl_linint2points.py` example script. This PR should be good now for approval and merge (unless there is anything else)

erogluorhan:
Thanks a lot! Comments helped me out much on improving things.

PR #65, opened 2020-04-20
month_to_season

### LINKED ISSUE #7: month_to_season
Port the NCL **month_to_season** family of functions, including: Month_to_season, month_to_seasonN, month_to_season12. The **month_to_season** function computes a user-specified three-month seasonal mean (DJF, JFM, FMA, MAM, AMJ, MJJ, JJA, JAS, ASO, SON, OND, NDJ).

bonnland:
CAPSTONE-118  Initial commit for Draft Pull Request

Closes #7


bonnland:
@khallock @jclyne This is ready for review.   I'm not expecting I have matched all naming conventions, but I tried.

src/geocat/comp/month_to_season.py
```diff
@@ -0,0 +1,39 @@
...
+        Time stamps are centered on the season. For example, seasons='DJF' returns January timestamps.
+
+        If a calculated season's timestamp falls outside the original range of monthly values, then the calculated mean
+        is dropped.  For example, if the monthly data's time range is [Jan-2000, Dec-2003] and the season is "DJF", the
+        seasonal mean computed from the single month of Dec-2003 is dropped.
+    """
+    mod_check(xMon.time.size, 12)
+
+    startDate = xMon.time[0]
+    endDate = xMon.time[-1]
```
andersy005, line 18: The time coordinate may not be called `time` for all dataset/dataarray objects. For instance, for some WRF output files it's called `Times`. So, for generality, we may want to have this as a configurable parameter. 

```python
def month_to_season(xMon, season, time_coord_nam='time'):
```

```python
startDate = xMon[time_coord_name][0]
etc....
```

src/geocat/comp/month_to_season.py
```diff
@@ -0,0 +1,39 @@
...
+    seasons_pd = {'DJF': ('QS-DEC', 1), 'JFM': ('QS-JAN', 2), 'FMA': ('QS-FEB', 3), 'MAM': ('QS-MAR', 4),
+                  'AMJ': ('QS-APR', 5), 'MJJ': ('QS-MAY', 6), 'JJA': ('QS-JUN', 7), 'JAS': ('QS-JUL', 8),
+                  'ASO': ('QS-AUG', 9), 'SON': ('QS-SEP', 10), 'OND': ('QS-OCT', 11), 'NDJ': ('QS-NOV', 12)}
+    try:
+        (season_pd, season_sel) = seasons_pd[season]
+    except KeyError:
+        raise ValueError("contributed: month_to_season: bad season: SEASON = " + season)
```
andersy005, line 25:
```suggestion
        raise ValueError(f"contributed: month_to_season: bad season: SEASON = {season}. Valid seasons include: {list(season_pd.keys())}")
```

src/geocat/comp/month_to_season.py
```diff
@@ -0,0 +1,39 @@
+
+
+def month_to_season(xMon, season):
+    """ This function takes an xarray dataset containing monthly data spanning years and
+        returns a dataset with one sample per year, for a specified three-month season.
```
andersy005, line 5: The docstring mentions that the expected input is `monthly_data`, what happens if the user provides **daily** data for example? Does this fail or raise an error?

src/geocat/comp/month_to_season.py
bonnland: Curious if this will upset anyone not using the latest python (3.8?).   Otherwise I think it is a good idea.

src/geocat/comp/month_to_season.py
bonnland: The only real check in the NCL version was that the number of time points is divisible by 12.   Practically speaking, this appears to have been fine for those using the function, but as an exercise I may add the check for monthly data.    

The best approach I have come up with so far, to decide if the data is monthly, is to create a Pandas monthly series from the start date to the end date and see if the number of points matches.  I am expecting the number might be "off by one", but maybe this is close enough to call it monthly data.

src/geocat/comp/month_to_season.py
bonnland: Sounds good.  I will add this.

src/geocat/comp/month_to_season.py
andersy005: Python f-strings are available from Python version 3.6+. So, this should not be a problem unless geocat supports earlier versions of Python.

src/geocat/comp/month_to_season.py
andersy005: >. The only real check in the NCL version was that the number of time points is divisible by 12. 

Is this check needed in xarray? I would say no. As long as the time axis is decoded, we can extract all date-time properties (month, day, quarter, etc....). Also, isn't it the case that checking if the time axis divisible by 12 is too fragile? For instance, with `calendar='360 days'`, a dataset with daily data with 360 days will pass the `divisible by 12` check even though it is not monthly data.

src/geocat/comp/month_to_season.py
bonnland: @jclyne @khallock   Has there been any discussion around which versions of python will be required to use geocat-comp?   We're wondering if using a feature introduced in python 3.6 is OK or not.

src/geocat/comp/month_to_season.py
andersy005: >. The best approach I have come up with so far, to decide if the data is monthly, is to create a Pandas monthly series from the start date to the end date and see if the number of points matches.

This approach assumes that we have data on a standard/Gregorian calendar. It will fail if the data is using other calendars(`360_day`, `no_leap`, `all_leap`).

bonnland: > Is this check needed in xarray? I would say no.

It is difficult to know the original intent of this check.  It seems a quick and dirty method for checking for monthly data.   If we can come up with a better check that works across all calendars, then maybe we should replace this check.   I am open to suggestions on a good way to check across all calendar types.

bonnland: > It will fail if the data is using other calendars(`360_day`, `no_leap`, `all_leap`).

I thought it might still work in most cases.   Trying to handle all possible cases, like a time range including the year 1582 when the Gregorian calendar was instituted and ten days were lost, seems unnecessary.

bonnland: Possibly you were thinking about daily data not working across calendars, instead of monthly?   Since the number of months is the same in all calendars?

andersy005: >. Possibly you were thinking about daily data not working across calendars, instead of monthly?

Both. 

1. If the function is exclusively for `monthly_data`, then we should have a check for time frequency maybe instead of relying on the size of the time axis?

2. Depending on how (1.) is implemented, we may not have to worry about the difference in calendars.

bonnland: >     1. If the function is exclusively for `monthly_data`, then we should have a check for time frequency maybe instead of relying on the size of the time axis?

This sounds like the right idea.  I could use some help understanding the full scope of time in xarray, so I can be sure this is done in a general enough way.

bonnland: I think in the original check was also the notion that there should be no gaps in the monthly data.   I don't know of an easy way to check this yet.

bonnland: In the interest of being conservative, I think it's OK to improve the monthly data check in the next version.

bonnland:
@erogluorhan I seem to be unable to select reviewers for this pull request.   Is someone available from the GeoCat team to review?    I am curious about whether there is explicit desire to support older versions of python (before version 3.6).  

I am also wondering if the CircleCI checks need some attention? Thank you!

src/geocat/comp/month_to_season.py
```diff
@@ -0,0 +1,39 @@
...
+                  'AMJ': ('QS-APR', 5), 'MJJ': ('QS-MAY', 6), 'JJA': ('QS-JUN', 7), 'JAS': ('QS-JUL', 8),
+                  'ASO': ('QS-AUG', 9), 'SON': ('QS-SEP', 10), 'OND': ('QS-OCT', 11), 'NDJ': ('QS-NOV', 12)}
+    try:
+        (season_pd, season_sel) = seasons_pd[season]
+    except KeyError:
+        raise ValueError("contributed: month_to_season: bad season: SEASON = " + season)
+
+    # Compute the three-month means, moving time labels ahead to the middle month.
+    month_offset = 'MS'
+    xSeasons = xMon.resample(time=season_pd, loffset=month_offset).mean()
```
andersy005, line 29:
```suggestion
    xSeasons = xMon.resample({time_coord_name: season_pd}, loffset=month_offset).mean()
```

src/geocat/comp/month_to_season.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Filter just the desired season, and trim to the desired time range.
+    xSea = xSeasons.sel(time=xSeasons.time.dt.month == season_sel)
+    xSea = xSea.sel(time=slice(startDate, endDate))
```
andersy005, line 33:
```suggestion
    xSea = xSeasons.sel({time_coord_name: xSeasons.time.dt.month == season_sel})
    xSea = xSea.sel({time_coord_name: slice(startDate, endDate)})
```

src/geocat/comp/month_to_season.py
bonnland: @andersy005 The time coordinate generalization is failing on this line:  

`xSeasons = xMon.resample(time=season_pd, loffset=month_offset).mean()`

It seems this function expects the time dimension is called "time".   Is there some way to communicate the proper time coordinate to the resample()?   I have done my best to look through the xarray documentation...

src/geocat/comp/month_to_season.py
bonnland: Aha, I see you addressed my question here.   Thank you!

src/geocat/comp/month_to_season.py
```diff
@@ -0,0 +1,39 @@
...
+                  'AMJ': ('QS-APR', 5), 'MJJ': ('QS-MAY', 6), 'JJA': ('QS-JUN', 7), 'JAS': ('QS-JUL', 8),
+                  'ASO': ('QS-AUG', 9), 'SON': ('QS-SEP', 10), 'OND': ('QS-OCT', 11), 'NDJ': ('QS-NOV', 12)}
+    try:
+        (season_pd, season_sel) = seasons_pd[season]
+    except KeyError:
+        raise ValueError("contributed: month_to_season: bad season: SEASON = " + season)
+
+    # Compute the three-month means, moving time labels ahead to the middle month.
+    month_offset = 'MS'
+    xSeasons = xMon.resample({time_coord_name: season_pd}, loffset=month_offset).mean()
```
bonnland, line 29: This line is still generating an error.   Did I miss something? 

```
test/test_month_to_season.py:88 (Test_month_to_season.test_custom_time_coordinate)
self = <test_month_to_season.Test_month_to_season testMethod=test_custom_time_coordinate>

    def test_custom_time_coordinate(self):
>       season_ds = month_to_season(self.custom_time_dataset, 'JFM', time_coord_name='my_time')

test/test_month_to_season.py:90: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../miniconda3/envs/geocat_build/lib/python3.7/site-packages/geocat/comp/month_to_season.py:29: in month_to_season
    xSeasons = xMon.resample({time_coord_name: season_pd}, loffset=month_offset).mean()
../../miniconda3/envs/geocat_build/lib/python3.7/site-packages/xarray/core/common.py:1100: in resample
    if isinstance(self.indexes[dim_name], CFTimeIndex):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = lat: Float64Index([-90.0], dtype='float64', name='lat')
lon: Float64Index([-180.0], dtype='float64', name='lon')
key = 'my_time'

    def __getitem__(self, key):
>       return self._indexes[key]
E       KeyError: 'my_time'

../../miniconda3/envs/geocat_build/lib/python3.7/site-packages/xarray/core/indexes.py:59: KeyError
```

bonnland: The other tests are passing, so it would seem that somewhere in xarray, the assumption of time='time' is still there (the other tests assume this also).

bonnland: Looking more closely, it appears that xarray has both self._indexes[] and self.indexes[], and these are being used interchangeably.

bonnland: If this is a bug, I would like the practice of opening the xarray issue.   I just can't tell right away if I missed something, or it really is a bug.

src/geocat/comp/month_to_season.py
clyne: That's a question for @khallock, but looping in @pilotchute and @erogluorhan as well.

test/test_month_to_season.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Create a dataset that combines the two previous datasets, for two years of data.
+        self.ds3 = xr.concat([self.ds1, self.ds2], dim='time')
+
+        # Create a dataset with the wrong number of months.
+        self.partial_year_dataset = get_fake_dataset(start_month='2000-01', nmonths=13, nlats=1, nlons=1)
+
+        # Create a dataset with a custom time coordinate.
+        custom_time_dataset = get_fake_dataset(start_month='2000-01', nmonths=12, nlats=1, nlons=1)
+        self.custom_time_dataset = custom_time_dataset.rename_dims({'time': 'my_time'})
```
andersy005, line 48:
```suggestion
        self.custom_time_dataset = custom_time_dataset.rename({'time': 'my_time'})
```

This is what is causing the test to fail. As the name hints, this will rename dimension names only. What you want is to rename the time coordinate as well. 

```python
In [20]: ds.rename_dims({'time': 'my_time'})
Out[20]:
<xarray.Dataset>
Dimensions:  (lat: 1, lon: 1, my_time: 12)
Coordinates:
    time     (my_time) datetime64[ns] 2000-01-01 2000-02-01 ... 2000-12-01
  * lat      (lat) float32 -90.0
  * lon      (lon) float32 -180.0
Dimensions without coordinates: my_time
Data variables:
    my_var   (my_time, lat, lon) float32 1.0 2.0 3.0 4.0 ... 9.0 10.0 11.0 12.0
```

```python
In [21]: ds.rename({'time': 'my_time'})
Out[21]:
<xarray.Dataset>
Dimensions:  (lat: 1, lon: 1, my_time: 12)
Coordinates:
  * my_time  (my_time) datetime64[ns] 2000-01-01 2000-02-01 ... 2000-12-01
  * lat      (lat) float32 -90.0
  * lon      (lon) float32 -180.0
Data variables:
    my_var   (my_time, lat, lon) float32 1.0 2.0 3.0 4.0 ... 9.0 10.0 11.0 12.0
```

src/geocat/comp/month_to_season.py
andersy005: See my comment below.

test/test_month_to_season.py
bonnland: Thank you again, Anderson!   Unit tests are all passing now.

src/geocat/comp/month_to_season.py
erogluorhan: Hi @bonnland we are currently providing continuous integration tests in Python 3.6, 3.7, and 3.8; thus, there should not be any problem with 3.6 as long as the PR passes all of them.

erogluorhan:
> @erogluorhan I seem to be unable to select reviewers for this pull request. Is someone available from the GeoCat team to review? I am curious about whether there is explicit desire to support older versions of python (before version 3.6).
> 
> I am also wondering if the CircleCI checks need some attention? Thank you!

Hi @bonnland, @clyne and I are now the reviewers of this PR. We already provide support for version 3.6 and upper, but versions before 3.6 would be of discussion, too, if there is any particular need or use case for it.

My understanding of the question about CircleCI is that if we should be concerned as current CircleCI tests are Required but "**Waiting for status to be reported**" (Please let me know if my understanding is incorrect): Yes, we require all four CircleCI tests to pass before merging a PR into `develop` branch. CircleCI tests of this PR seem to be waiting likely due to **branch protection rules**. I could not figure out the reason for it yet as your branch seems up to date with the upstream but still looking into it. May @khallock have any idea on it?

src/geocat/comp/month_to_season.py
bonnland: Marking as resolved for now, but will reconsider if users complain.   In practice, daily data is likely to fail the "time points divisible by 12" check, but not guaranteed.

src/geocat/comp/month_to_season.py
bonnland: Committed.

bonnland:
@erogluorhan @khallock    My GitHub actions were enabled already, and it is my guess that CircleCI is supposed to be running in the NCAR repository for this pull request, not my fork.    Is there a GitHub App or Auth Token that I need for permission to build?   CircleCI is not letting me add or follow this repo because I do not have "Organization Admin" permissions in Github for the NCAR organization.

khallock:
I figured out why this isn't triggering builds, apparently there's a separate setting on CircleCI for enabling builds for PRs based on forks: https://circleci.com/gh/NCAR/geocat-comp/edit#advanced-settings

I think this should be resolved when a new commit is pushed, so I'll try re-pushing the latest commit and see if that does the trick.

clyne:
Awesome. Thanks @khallock. Please let us know the results.

khallock:
It seems to have worked, the tests are passing now!

src/geocat/comp/month_to_season.py
```diff
@@ -0,0 +1,39 @@
...
+
+    startDate = xMon[time_coord_name][0]
+    endDate = xMon[time_coord_name][-1]
+    seasons_pd = {'DJF': ('QS-DEC', 1), 'JFM': ('QS-JAN', 2), 'FMA': ('QS-FEB', 3), 'MAM': ('QS-MAR', 4),
+                  'AMJ': ('QS-APR', 5), 'MJJ': ('QS-MAY', 6), 'JJA': ('QS-JUN', 7), 'JAS': ('QS-JUL', 8),
+                  'ASO': ('QS-AUG', 9), 'SON': ('QS-SEP', 10), 'OND': ('QS-OCT', 11), 'NDJ': ('QS-NOV', 12)}
+    try:
+        (season_pd, season_sel) = seasons_pd[season]
+    except KeyError:
+        raise ValueError(f"contributed: month_to_season: bad season: SEASON = {season}. Valid seasons include: {list(seasons_pd.keys())}")
```
andersy005, line 25: For consistency, you may want to raise a `KeyError` here instead of a `ValueError`

```suggestion
    except KeyError:
        raise KeyError(f"contributed: month_to_season: bad season: SEASON = {season}. Valid seasons include: {list(seasons_pd.keys())}")
```

bonnland: Thanks Anderson, good suggestion.   The original thinking was that the user doesn't need to know the internals of this function, like that this string was used as a key in a dictionary.   But I sense that the python community worries less about hiding implementations than other languages.   So it makes sense to call it what it is.

andersy005:
@bonnland, don't forget to update your tests. some tests are expecting `ValueError`.

bonnland:
A disadvantage of using GitHub to commit code; I should have checked first.

bonnland:
@khallock  CircleCI ncomp build is failing to create an environment, with this error message:  

`error: pathspec 'pull/65' did not match any file(s) known to git`   

It looks like some packages in the environment are inconsistent with each other, for some reason.

build_envs/environment_Linux.yml
```diff
@@ -12,3 +12,4 @@ dependencies:
   - distributed
   - pytest
   - ncomp
+  - netcdf4
```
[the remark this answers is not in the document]
bonnland: Hi Orhan,  the netcdf4 library is not used in my code submission, but it is used when building the environment on CircleCI in a file called `run_checks.sh`.   I do not know the purpose of this file, but it seems to install seaborn and netcdf4.  

Is anyone on the GeoCAT team aware of what `run_checks.sh` is for?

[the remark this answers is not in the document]
erogluorhan: Hi Brian, I checked the history of that file (`run_checks.sh`), and the addition of **netcdf4** and **scipy** dependencies seem to have been done well before our [geocat-comp](https://github.com/NCAR/geocat-comp) / [geocat-ncomp](https://github.com/NCAR/geocat-ncomp) repository separation. These dependencies are indeed required for the test of a function that is now belonging to **geocat-ncomp**, i.e. not this repository. I have also already checked in another PR that removing these dependencencies should not hurt this repository. If that's also the case for you, i.e. if you haven't observed any error before adding **netcdf** dependency to environment_<platform>.yml, could you please remove it? 

FYI: We are still dealing with the overall dependencies between geocat-comp and geocat-ncomp with Austin and will handle dependency-related issues / decisions in a different work soon.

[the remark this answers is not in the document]
bonnland: OK, thank you for that clarification.   I am curious to see if the CircleCI checks will pass.

erogluorhan:
They have passed all the checks now!

bonnland:
@erogluorhan   Thank you!   Letting you and the team continue from here.

clyne:
@pilotchute please review when you can. Thanks!

src/geocat/comp/month_to_season.py
```diff
@@ -0,0 +1,39 @@
...
+
+        This function requires the number of months to be a multiple of 12, i.e. full years must be provided.
+
+        Time stamps are centered on the season. For example, seasons='DJF' returns January timestamps.
+
+        If a calculated season's timestamp falls outside the original range of monthly values, then the calculated mean
+        is dropped.  For example, if the monthly data's time range is [Jan-2000, Dec-2003] and the season is "DJF", the
+        seasonal mean computed from the single month of Dec-2003 is dropped.
+    """
+    mod_check(xMon[time_coord_name].size, 12)
```
pilotchute, line 15: This mod_check function is only used once, Would you want to replace the function call with the function contents here, or move mod_check into a helper function library file? 

This isn't a big issue, but will affect maintainability and as such is a code style change that should be addressed.

bonnland: It's a good and reasonable question.   The motivation behind `mod_check()` as a function came from the original NCL code base, where it was also a function.   

IIRC, many user-contributed NCL functions, such as `month_to_season()` were merged into a single file, and I originally assumed that `mod_check()` might be called by more than one function.   If that's the case, then there would be some support for keeping `mod_check()` around.   Otherwise, I would agree that the code be merged into the calling function.

bonnland: I'm seeing some evidence that the NCL version of the method, called `modCheck()`, was called by other user-contributed functions: 

https://sites.google.com/site/earthsciencedataanalysis/software-tools-list/chapters/classifications/data-analysis-tools/ncl/ncl-script/NCL_Functions/buildseasonfrommonthlydata

So if the GeoCAT team wants to keep `mod_check` as a good starting helper function, or has already designated a helper function library location,  I am happy to move the function.

bonnland:
> I appreciate your help in learning a pattern for handling merge requests, and would like to talk to you about what levels of code style strictness you as a contributor would be willing to work with.

I'm generally a fan of PEP8, though still getting used to writing code that meets all the requirements.  I have been using the pre-commit library for auto-correcting code at commit time, which is convenient but doesn't necessarily give me feedback on what was corrected.  So I am maybe learning more slowly with pre-commit fixing my mistakes.

pilotchute:
Once #79 is merged, I'll merge this branch.

src/geocat/comp/month_to_season.py
pilotchute: We'll keep it in month_to_season for now, thanks for providing the background.

pilotchute: When we implement a helper function library, we'll likely move it then. 
I'm running the autopep8 formatter on your branch now, and I'll merge once the checks complete.

PR #67, opened 2020-04-27
Remove geocat-ncomp functionality

khallock:
Removed Cython code and the Python functions wrapping it, migrated functionality into new [geocat-ncomp](https://github.com/NCAR/geocat-ncomp) package. If geocat-ncomp is installed, then geocat-comp imports all of its functions into the `geocat.comp` namespace using `from geocat.ncomp import *`.

The geocat-comp documentation still contains references to the top-level Python functions; for example, this means that `geocat.comp.linint2` will have a documentation page on both the geocat-comp and geocat-ncomp documentation, while the Cython function `geocat.ncomp._linint2` will only have documentation on the geocat-ncomp site.


khallock:
Docs built from this branch: https://geocat-comp.readthedocs.io/en/remove_geocat-ncomp/index.html

Issue #62, opened 2020-04-07
Docstring refactoring


erogluorhan:
All of the functions should be refactored with respect to following points:

1. Python API: Consistent language use and descriptions for function arguments should be achieved in the docstrings.

2. Cython: Most of the docstrings in this level are duplicates with Python API docstrings. To prevent such duplication, repetitive information in this level may be removed. In other words, only different information than Python API may be kept in this level.

pilotchute:
Docstrings will be updated in polynomial.py and then once the ncomp migration is done, a separate branch will be made for ncomp cython docstring changes. 

Docstrings will comply with [PEP 257](https://www.python.org/dev/peps/pep-0257/)

pilotchute:
The Sphinx documentation system does weird things with certain keywords "Args" "Parameters" "Returns"

I'm going to explore a few ways of formatting the docstrings such that those keywords are avoided so that formatting can be controlled a bit more.

erogluorhan:
Hi @pilotchute and @clyne 

I think this issue has been automatically closed with closing of PR #70, but I think we need to reopen it due to the existing need for addressing the same issue for Python and Cython levels at geocat-ncomp. Please correct me and re-close it if I am wrong.

clyne:
Hi @erogluorhan 

Just to be clear are you saying we need to perform the same refactoring in geocat-ncomp? It was my understanding that geocat-ncomp contained only Fortran and Cython code. Is there duplication in the documentation between Fortran and Cython, lack of standardization, or both?

erogluorhan:
> "It was my understanding that geocat-ncomp contained only Fortran and Cython code."

Actually, whenever we have a Python code that has Fortran dependency and thus Cython dependency, that code goes into **geocat-ncomp** with its Cython code, e.g. `linint2`, `rcm2rgrid`, `linint2_points`, `eof`, etc. From this point, **geocat-ncomp** contains Python and Cython code only. In addition, **libncomp**, contains corresponding C and Fortran codes for those in **geocat-ncomp**.  On the other hand, whenever we have pure Python code (without any Fortran dependency), that goes into **geocat-comp** indeed, e.g. polynomial.py etc. 

> "Just to be clear are you saying we need to perform the same refactoring in geocat-ncomp? Is there duplication in the documentation between Fortran and Cython, lack of standardization, or both?"

[...]

PR #70, opened 2020-05-04
Docstring refactoring

pilotchute:
Starting a draft pull request to link against #62


src/geocat/comp/polynomial.py
```diff
@@ -65,55 +67,55 @@ def _unchunk_ifneeded(data: da.Array, axis: int) -> da.Array:
...
 
-    Args:
+    Parameters
+    ----------
 
-        x (:class:`array_like`):
-            x-coordinate, an Iterable object of shape `(M,)`, `(M, 1)`, or `(1, M)` where `M = y.shape(axis)`. It cannot
-            have `nan` or missing values.
+    x : :class:`array_like`
+        X-coordinate, an iterable object of shape `(M,)`, `(M, 1)`, or `(1, M)` where `M = y.shape(axis)`. It cannot
```
khallock, line 76: An `Iterable` is a Python class, I would consider keeping the 'I' capitalized for clarity.

src/geocat/comp/polynomial.py
```diff
@@ -65,55 +67,55 @@ def _unchunk_ifneeded(data: da.Array, axis: int) -> da.Array:
...
+        have `nan` or missing values.
 
-        y (:class:`array_like`)
-            y-coordinate, an Iterable containing the data. It could be list, `numpy.ndarray`, `xr.DataArray`, Dask array.
-            or any Iterable convertible to `numpy.ndarray`. In case of Dask Array, The data could be chunked. It is
-            recommended no to chunk along the `axis` provided.
+    y : :class:`array_like`:
+        Y-coordinate, an iterable containing the data. It could be list, `numpy.ndarray`, `xr.DataArray`, Dask array.
```
khallock, line 80: See above.

docs/api.rst
```diff
@@ -6,3 +6,4 @@ API Reference
    :maxdepth: 2
    
    user_api/index.rst
+   internal_api/index.rst
```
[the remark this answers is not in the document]
pilotchute: My personal preference is for the internal API to be visible, since only it's use is intended to be internal, but there is no reason to restrict information about it.

PR #75, opened 2020-06-13
Swap ncomp for geocat ncomp

pilotchute:
large number of commits because I rebased against develop after I committed the version bump.


pilotchute:
once this version bump is on develop, I can publish the new packages to the NCAR conda channel

PR #79, opened 2020-06-24
github autopep8 action on push and pull_request

pilotchute:


pilotchute:
@clyne This is set up to autoformat code on a push in this order:
User push to repo triggers the autopep8 workflow, which checks out the pushed head on an unbuntu docker instance, and runs the [autopep8](https://github.com/hhatto/autopep8) program on checked out repo, then creates an automatic push to add the formatting changes to the branch.

pilotchute:
The second push only happens if autopep8 makes any changes. So if autopep8 is run locally before it a commit is pushed, no second automated push will occur.

Issue #91, opened 2020-12-08
Implementation of relhum


anissa111:
This issue is a part of a larger effort detailed in this [S@S issue](https://github.com/NCAR/science-at-scale/issues/1).

PR #93, opened 2021-01-04
Relhum

anissa111:
Addresses #91 

This is currently a work in progress as dask and xarray support is added for the purpose of supporting S@S efforts.

Summary:
 - implementation of relhum
 - basic testing added using ground truth from ncl
 - support for numpy


src/geocat/comp/relhum.py
```diff
@@ -0,0 +1,119 @@
...
+    # Convert inputs to np arrays
+    temperature = np.asarray(temperature)
+    mixing_ratio = np.asarray(mixing_ratio)
+    pressure = np.asarray(pressure)
+
+    # make an empty space for output array
+    relative_humidity = np.zeros(np.size(temperature))
+
+    # fill in output array
+    for i in range(np.size(temperature)):
```
ahuang11, line 42: I don't think the loop and the flatten/ravel is necessary here.
```
    # Compute directly
    relative_humidity = _relhum_tdd(
        temperature,
        mixing_ratio,
        pressure
    )
```
Loops' runtimes scale linearly, e.g. 10,000 grid points takes ~10x longer than 1000 grid points, which isn't ideal if computing relative humidity on a fine-resolution grid.

anissa111: Hello! Sorry for the late reply to your points. You are absolutely correct, and this set up was only intended to be temporary and has been altered in the final version of the PR.

ahuang11: No worries, just offering unsolicited advice~ hope it helps!

src/geocat/comp/relhum.py
```diff
@@ -0,0 +1,205 @@
...
+                Pressure in Pa. Must have the same dimensions as temperature
+
+            Returns
+            -------
+            relative_humidity : numpy.ndarray or float
+                Relative humidity. Must have the same dimensions as temperature
+        """
+
+    # make sure the input arrays are of the same size
+    if np.shape(temperature) != np.shape(mixing_ratio) or np.shape(
```
ahuang11, line 27: Since you used np.shape(temperature) 3x, maybe save to variable

src/geocat/comp/relhum.py
```diff
@@ -0,0 +1,205 @@
...
+            f"dewtemp_trh: dimensions of temperature, {np.shape(temperature)}, and mixing ratio, "
+        )
+
+    # check if single value input, and skip dask if appropriate
+    if np.size(temperature) == 1:
+        return _relhum_tdd(temperature, mixing_ratio, pressure)
+
+    # ''' Start of boilerplate
+    if not isinstance(temperature, xr.DataArray):
+        temperature = xr.DataArray(temperature)
```
ahuang11, line 39: Maybe you don't need to cast to xr.DataArray. da.from_array accepts numpy, unless you want to handle lists / pd.Series, np.asarray might be a better candidate.

Also, if you do cast to xr.DataArray, xarray has its own map blocks
https://xarray.pydata.org/en/stable/generated/xarray.map_blocks.html#xarray-map-blocks

src/geocat/comp/relhum.py
```diff
@@ -0,0 +1,205 @@
...
+    # Define data variables
+
+    t0 = 273.15
+    ep = 0.622
+    onemep = 0.378
+    es0 = 6.1128
+    a = 17.269
+    b = 35.86
+
+    est = es0 * exp((a * (t - t0)) / ((t - t0) - b))
```
ahuang11, line 200: If an array, maybe use np.exp instead of math.exp

src/geocat/comp/relhum.py
anissa111: Thank you for the review, the way this is handled has been updated

src/geocat/comp/relhum.py
anissa111: addressed, thank you

erogluorhan (approved):

Issue #94, opened 2021-01-04
implementation of dewtemp


anissa111:
This issue is a part of a larger effort detailed in this [S@S issue](https://github.com/NCAR/science-at-scale/issues/1).

PR #95, opened 2021-01-04
Dewtemp

anissa111:
Addresses #94 

This is currently a work in progress as dask and xarray support is added for the purpose of supporting S@S efforts.

Summary:
- implementation of dewtemp
- basic testing added using ground truth from ncl
- support for numpy


src/geocat/comp/dewtemp_trh.py
```diff
@@ -0,0 +1,76 @@
+import numpy as np
+import xarray as xr
+import dask.array as da
+from dask.array.core import map_blocks
+
+
+def dewtemp_trh(temperature, relative_humidity):
+    """ This function calculates the dew point temperature given temperature and relative humidity
```
erogluorhan, line 8: For docstrings, I would suggest using the format (i.e. indentation, keywords such as Args and Return, etc.) from geocat-f2py (e.g. dpres_plevel_wrapper.py) because it was decided to be most favorable based on an investigation.

test/test_dewtemp_trh.py
```diff
@@ -0,0 +1,36 @@
...
+dt_td_1 = 6.3
+
+dt_td_2 = [
+    24.38342, 19.55563, 15.53281, 16.64218, 16.81433, 14.22482, 9.401337,
+    6.149719, -4.1604, -5.096619, -6.528168, -12.61957, -19.38332, -25.00714,
+    -28.9841, -33.34853, -46.51273, -58.18289
+]
+
+
+class Ttest_relhum(unittest.TestCase):
```
erogluorhan, line 16: Please add two extra test cases:
1) An additional array-run-test that tests for Xarray inputs rather than Numpy.
2) An additional array-run-test that tests for an already-chunked array. It may be a Dask array because you seem to already be converting Numpy inputs to first Xarray then Dask array.

src/geocat/comp/dewtemp_trh.py
```diff
@@ -0,0 +1,76 @@
...
+    # ''' Start of boilerplate
+    if not isinstance(temperature, xr.DataArray):
+        temperature = xr.DataArray(temperature)
+        temperature = da.from_array(temperature, chunks="auto")
+
+    if not isinstance(relative_humidity, xr.DataArray):
+        relative_humidity = xr.DataArray(relative_humidity)
+        relative_humidity = da.from_array(relative_humidity, chunks="auto")
+
+    tdk = map_blocks(_dewtemp, temperature, relative_humidity)
```
erogluorhan, line 44: My understanding from this entire function is that if the input is Numpy, the code is converting it to first an Xarray, and then to an auto-chunked Dask array. However, if it is an Xarray (either chunked or unchunked), then it does not perform any chunking or conversion to Dask array. Alternatively, the code seems to be capable of taking inputs as already chunked Dask arrays, but this is not documented. So, please test those Xarray and Dask input cases in unit tests (for which I have already put a comment below), and document input alternatives accordingly in the docstrings.

erogluorhan (changes requested):
The code looks good to me, and I can confirm tests are passing. Please address a few inline comments.

src/geocat/comp/dewtemp.py
```diff
@@ -0,0 +1,103 @@
...
+    # Call mapblocks to run function
+    dew_pnt_temp = map_blocks(_dewtemp, temperature, relative_humidity)
+    dew_pnt_temp = dew_pnt_temp.compute()
+
+    # tdk = _dewtemp(temperature, relative_humidity)
+    return dew_pnt_temp
+
+
+# make alias for original function name
+def dewtemp_trh(temperature, relative_humidity):
```
erogluorhan, line 56: What purpose is this alias for? Can't we choose either name and go with that one in the user API?

src/geocat/comp/dewtemp.py
```diff
@@ -0,0 +1,103 @@
...
+
+                dew_pnt_temp (:class:`numpy.ndarray`, :class:`xr.DataArray`, :obj:`list`, or :obj:`float`):
+                    Dewpoint temperature in Kelvin. Same size as input variable temperature
+    """
+
+    return dewtemp(temperature, relative_humidity)
+
+
+def _dewtemp(tk, rh):
+    """ This function calculates the dew point temperature given temperature and relative humidity
```
erogluorhan, line 79: I am not sure if we need this docstrings, which is very close to that of the user API function other than argument names.

src/geocat/comp/dewtemp.py
```diff
@@ -0,0 +1,103 @@
+import numpy as np
+import xarray as xr
+import dask.array as da
+from dask.array.core import map_blocks
+
+
+def dewtemp(temperature, relative_humidity):
+    """ This function calculates the dew point temperature given temperature and relative humidity
```
erogluorhan, line 8: I am sorry to realize it now but I misled you to this docstrings format. The format of our preference is the Google format, which would make this docstrings look like (It is not any issue but would be better to be consistent with one format of choice and apply it to our new PRs like here):

```
"""
    This function calculates the dew point temperature given temperature and relative humidity
    using equations from John Dutton's "Ceaseless Wind" (pp 273-274)

    Parameters
    ----------

    temperature : :class:`numpy.ndarray`, :class:`xr.DataArray`, :obj:`list`, or :obj:`float`:
        Temperature in K

    relative_humidity : :class:`numpy.ndarray`, :class:`xr.DataArray`, :obj:`list`, or :obj:`float`:
        Relative humidity. Must be the same dimensions as temperature


    Returns
    -------

    dew_pnt_temp : :class:`numpy.ndarray` or :obj:`float`:
        Dewpoint temperature in Kelvin. Same size as input variable temperature
"""
```

erogluorhan (commented):
Looks great overall; please address a few inline comments, and then we are done

src/geocat/comp/dewtemp.py
```diff
@@ -0,0 +1,103 @@
...
+    if not isinstance(relative_humidity, xr.DataArray):
+        relative_humidity = xr.DataArray(relative_humidity)
+
+    # Make sure dask arrays are autochunked dask arrays
+    relative_humidity = da.from_array(relative_humidity, chunks="auto")
+    temperature = da.from_array(temperature, chunks="auto")
+
+    # Call mapblocks to run function
+    dew_pnt_temp = map_blocks(_dewtemp, temperature, relative_humidity)
+    dew_pnt_temp = dew_pnt_temp.compute()
```
andersy005, line 49: 👋🏽  @anissa111, I have a couple questions/comments:

1) Does this computation have to be done via dask? I'm imagining scenarios in which the user provides in-memory numpy arrays and their computations are routed through dask (which adds some performance overhead). They (user) will end up paying a price for using dask even when they don't want to use dask. 

2) It's my understanding that computation in `_dewtemp` consists of array elementwise operations and not blockwise (chunk) operations. I'm wondering whether the `map_blocks()` approach is needed here?

src/geocat/comp/dewtemp.py
```diff
@@ -0,0 +1,103 @@
...
+                dew_pnt_temp (:class:`numpy.ndarray` or :obj:`float`):
+                    Dewpoint temperature in Kelvin. Same size as input variable temperature
+    """
+
+    # make sure the input arrays are of the same size
+    if np.shape(temperature) != np.shape(relative_humidity):
+        raise ValueError(
+            f"dewtemp_trh: dimensions of temperature, {np.shape(temperature)}, and relative_humidity, "
+            f"{np.shape(relative_humidity)}, do not match")
+
```
andersy005, line 31: The docstring mentions that this function supports `xarray.DataArray`. However, this check appears not to handle xarray.DataArray objects correctly. xarray is more flexible than numpy when it comes to the shapes of arrays since it uses labelled axes instead of the plain location of the axis in the array. 

For instance, consider these two dataarrays of shapes `(3, 2)` and `(2, 3)`:


```python
In [1]: import xarray as xr

In [2]: import numpy as np

In [3]: a = xr.DataArray(np.ones(6).reshape(2, 3), dims=['lat', 'lon'])

In [4]: a
Out[4]:
<xarray.DataArray (lat: 2, lon: 3)>
array([[1., 1., 1.],
       [1., 1., 1.]])
Dimensions without coordinates: lat, lon

In [5]: b = a.transpose()

In [6]: b
Out[6]:
<xarray.DataArray (lon: 3, lat: 2)>
array([[1., 1.],
       [1., 1.],
       [1., 1.]])
Dimensions without coordinates: lon, lat

In [7]: b.shape
Out[7]: (3, 2)

In [9]: a.shape
Out[9]: (2, 3)
``` 


- Adding these two arrays (elementwise) provides the correct result since xarray is used the `lat` and `lon` axis labels to figure out whether the two arrays are compatible via the broadcasting rules:

```python
In [12]: a + b
Out[12]:
<xarray.DataArray (lat: 2, lon: 3)>
array([[2., 2., 2.],
       [2., 2., 2.]])
Dimensions without coordinates: lat, lon
```

- On the other hand, numpy fails because it only cares about the location of an axis in an array:
- 
```python
In [13]: a.data + b.data
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-13-1892839ecaba> in <module>
----> 1 a.data + b.data

ValueError: operands could not be broadcast together with shapes (2,3) (3,2)
```

I think that for xarray, this check should check whether the input arrays are [_broadcastable_](https://numpy.org/doc/stable/reference/ufuncs.html#broadcasting) or just try using the `xr.broadcast()` function to broadcast the two arrays against one other:

```python

In [17]: xr.broadcast(a, b)
Out[17]:
(<xarray.DataArray (lat: 2, lon: 3)>
 array([[1., 1., 1.],
        [1., 1., 1.]])
 Dimensions without coordinates: lat, lon,
 <xarray.DataArray (lat: 2, lon: 3)>
 array([[1., 1., 1.],
        [1., 1., 1.]])
 Dimensions without coordinates: lat, lon)
```

src/geocat/comp/dewtemp.py
andersy005: Presuming you are using the [numpydoc docstring guide](https://numpydoc.readthedocs.io/en/latest/format.html), I don't think the `:class:`, `:obj:` are needed:


- :class:`numpy.ndarray` - > numpy.ndarray
- :class:`xr.DataArray` -> xr.DataArray
- :obj:`list` -> list

etc....

andersy005: Also, according to https://www.python.org/dev/peps/pep-0257/, the indentation in https://github.com/NCAR/geocat-comp/pull/95#discussion_r574738745 shouldn't be there:

```python
"""Compute ..... 

Function to compute .... 
..... 

Parameters
---------
....

Returns
-------
....

"""

```

pilotchute (approved):
I think that once the other reviewer's comments are addressed, that this will be good to go. One stamp of approval.

src/geocat/comp/dewtemp.py
pilotchute: My understanding that without using dask, the elementwise operations will only use a single core, while the data could be chunked to allow for elementwise ops to happen in parallel.

src/geocat/comp/dewtemp.py
anissa111: Anderson, you are totally right about all of the above. The docstring format was copied from elsewhere internally, and unifying them will be handled after the first release.

erogluorhan (approved):

PR #96, opened 2021-01-05
skewt_params

michaelavs:
Code containing MetPy functions which would calculate parameters found in NCL skewt examples and display them below the plot title. **run with MetPy 1.0!!**
 
Currently, all but one function from metPy directly calculates the parameters of interest. Shox (showalter index) is not currently a function in MetPy but can be calculated by using the dry_lapse and moist_lapse functions to simulate lifting the parcel dry adiabatically from reference pressure (p[0]) to the lifting condensation level (roughly p[2] which is hard coded and will be resolved in future versions), then moved moist adiabatically from the lcl to 500mb (p[11], also hard coded but resolvable). This gives the temperature of the _parcel_. The final step is to subtract the parcel temp from _environmental_ temp. This will then give the Showalter index number. A brief overview of this equation can be found [here](http://www.zamg.ac.at/docu/Manual/SatManu/main.htm?/docu/Manual/SatManu/Basic/Convection/Parameters.htm) and an online calculator for the index is [here](https://www.weather.gov/epz/wxcalc_showalter). Both give a value of about -7 while the NCL version gives a value of 3. Both values are possible, however, we should not be seeing this large of a discrepancy between the NCL version and the Python version.


src/geocat/comp/skewt_params.py
```diff
@@ -0,0 +1,59 @@
...
+print("The parcel temp at 500mb is", tac[11])
+##############################################################################
+
+# CAPE and CIN
+cape = mpcalc.cape_cin(p, tc, tdc, ta)
+# print(cape[0].magnitude)
+
+# Precipitable Water
+precp = mpcalc.precipitable_water(p, tdc)
+precp = precp / 10  # Convert mm to cm
```
erogluorhan, line 38: This does not seem to be correct. Since `precp` is of type `Quantity`, I think the correct code should be as follows:

`precp = (precp.magnitude / 10) * units.centimeter`

In other words, the current version of the code results in precp.magnitudes divided by 10 however their unit to be kept as milimeter

Though, it is not directly related to the `shox` calculation (Even more `precp` is used nowhere in the code), but just wanted to correct the calculation.

src/geocat/comp/skewt_params.py
```diff
@@ -0,0 +1,59 @@
...
+# print(tc[11])
+
+tdc = (ds[9].values + 2) * units.degC  # Dew pt temp  [C]
+ta = mpcalc.parcel_profile(p, tc[0], tdc[0])  # Parcel profile
+tac = (ta.magnitude - 273.15) * units.degC  # Parcel temp in C
+print("The parcel temp at 500mb is", tac[11])
+##############################################################################
+
+# CAPE and CIN
+cape = mpcalc.cape_cin(p, tc, tdc, ta)
```
erogluorhan, line 33: Please note this `cape` value (2957.9725) also seems to be different than what NCL calculates (4455). 

Though, `cape` does not seem to be used in `shox` calculation if I am not wrong.

erogluorhan (commented):
Thank you very much for bringing this implementation up!

There are some details in the code that would need attention, which I pointed out in inline comments; however, they are not directly related to `shox` calculation.

I had a look at the online calculator you provided and it seems to be using the ml and ttop, which we can feed only from our Python code, i.e. not directly calculated in NCL code. This results in calculating -7 in our code and online calculator. 

I walked through our code and NCL code, and it looks like we are agreement on data `p`, `tc`, `tdc`. NCL code also inputs `z`, which is simply `ds[2].values` into the function, which may or may not have been used for `shox` calculation, I didn't have time to dig further on the algorithm.

Furthermore, `metpy.calc.lifted_index` looks pretty much similar to `shox` except that it uses a different initial location of the lifted parcel. However, it may be worth giving a look to see if this function could somehow be used for `shox` calculation.

I did could nit further work on the code but I am happy to make further work on it in the coming days. Please evaluate my comments and let me know if I can be of any help.

michaelavs:
@erogluorhan I created a function that manipulates the data and returns it as a string that can then be read in by the code and used as a subtitle on the diagram. The updated code uses the skewt_2_2 example code for validation of the function but if the PR is to be merged, it would be easy to remove the example code as a next step (there are also comments in the code that will be updated/removed as needed before merge). below is the original NCL version of the diagram followed by the version I have created using the function in python:

NCL 
image

Python
image

erogluorhan:
Thanks @michaelavs I wanted to have a look at it but the code seems to have been pushed with several "CONFLICT" indicators, e.g. `<<<<<<< HEAD`. Could you please clean them up and re-push?

michaelavs:
> Thanks @michaelavs I wanted to have a look at it but the code seems to have been pushed with several "CONFLICT" indicators, e.g. `<<<<<<< HEAD`. Could you please clean them up and re-push?

Sorry about that! Should be all fixed now!

erogluorhan:
@michaelavs I confirm that I now get the same `cape` value with you. The implementation of function itself looks good to me, and I agree with you that we can remove the demonstration codes when we think the function code is ready to be merged. I will walk through the code in detail once again to see why a significantly semantic difference between NCL-generated and our `shox` param (-7 vs 3). I will et you know if I can find anything.

erogluorhan (commented):
To investigate why our `shox` is different than NCL's calculation, you may want to have a look at NCL code here (If you haven't done so far):

`DSHOWALSKEWT` at Lines 214 and 296 of https://github.com/NCAR/ncl/blob/develop/ni/src/lib/nfpfort/skewT.f

and you may track how the parameters in `SHWLTR = T500 - TP500` get initialized and calculated. A potential difference in these parameters' calculation is probably the reason our result differs from NCL

michaelavs:
> To investigate why our `shox` is different than NCL's calculation, you may want to have a look at NCL code here (If you haven't done so far):
> 
> `DSHOWALSKEWT` at Lines 214 and 296 of https://github.com/NCAR/ncl/blob/develop/ni/src/lib/nfpfort/skewT.f
> 
> and you may track how the parameters in `SHWLTR = T500 - TP500` get initialized and calculated. A potential difference in these parameters' calculation is probably the reason our result differs from NCL

I think I found where the differences are between the two functions is coming from. It appears that in the NCL calculation, there are a lot of assumptions made about the heights of certain parameters. For example, on the NCL code line 270-271, the lcl is taken to be a standard 850 mb. To test this, I plugged in 850 in the Python code on line 93 which would read:
```
dl = mpcalc.dry_lapse(850*units.hPa, tac[0], p[0])
```
this changed the shox value to 8, still above the NCL version but definitely shows the impact of using a specific lcl level over calculated lcl. In the spirit of curiosity, I did find that the way to get the shox value to be 3 is by using the condition that the initial pressure is 900mb and using the coinciding 900mb temperature like this:
```
dl = mpcalc.dry_lapse(850*units.hPa, tac[3], p[3])
```
I will say, I don't know if this finding fully represents what the NCL code is doing as it would imply that the NCL code is using 900mb height instead of the 850mb or 800mb height in the code, but just something I thought was interesting and that shows a shox of 3 is possible with the dataset. 

With all of that being said, I think this opens up a couple of routes for this function. 
1) Either set some hard code values in the code to reflect the NCL style more closely. Downside is that assumptions of values could obviously lead to very skewed return values after manipulation.
2) Keep the python code as is, but maybe add some documentation for NCL-to-Python users that addresses this difference and why we kept it instead of doing a 1-to-1 match with NCL. Downside would be not having a true match in the Python code vs the NCL code.

erogluorhan:
Hi @michaelavs I noticed yapf formatter is just used in this PR. We removed them (yapf.yaml and build_test.yml) from `main` with PR #133 and are using `pre-commit hooks` and automated `ci.yml` from now on. You can do:

`git pull origin main`  (And resolve if any conflicts occur in your branch) to get rid of `yapf` and other older assets
`pre-commit run --all-files` to ensure the files that were already pushed in this PR is checked once by pre-commit

and then push your commits.

michaelavs:
@erogluorhan I went through and updated my branch as you had outlined, and it appears to be working. However, I'm not quite sure why the pre-commit check is failing for the last commit. Do you have a bit more insight on what I may be overlooking with this failure?

erogluorhan:
> @erogluorhan I went through and updated my branch as you had outlined, and it appears to be working. However, I'm not quite sure why the pre-commit check is failing for the last commit. Do you have a bit more insight on what I may be overlooking with this failure?

In your local branch, run

`pre-commit run --all-files`

It will first give you a report (starting with yapf) with one ore more "Fail"s, but at the same time it will have fixed them on the corresponding files. If you want to see in details which files were leading to those pre-commit errors and what lines in those files were the reasons, just check them:

`git status`
`git diff <file_name>`

When you are convinced about these pre-commit changes, you should not forget to commit these changes.

NOTE: After the first run of pre-commit and seeing a numebr of failures, when you run `pre-commit run --all-files` a second time, you should be able to see an all-passing report as below (And when you commit an all-passing report, then your PR should pass pre-commit checks in remote as well):

image

NOTE 2: In general, you may not always need to run `pre-commit run --all-files`. Instead, whatever the touched files you are attempting to commit, the pre-commit hooks catch them and automatically fixes them, giving you a report about failures or passes. You then shouldn't forget to add and commit those fixed files. The reason we are running pre-commit for all the files here is that you never ran it before and hadn;t puleld the `main` branch either.

Anyway, I hope this helps; let me know otherwise.

erogluorhan:
@michaelavs its tests look good now but I am not sure to give another review cycle as you haven't re-requested yet. Let me know.

michaelavs:
> @michaelavs its tests look good now but I am not sure to give another review cycle as you haven't re-requested yet. Let me know.

I just re-requested your review! 

Just in case you are wondering: once everything has been approved, I will take out the test data that is in the file, leaving behind lines 31-150. So anything that isn't directly related to the functions will be removed before merge.

erogluorhan (changes requested):
Could you please write unit tests for it?

erogluorhan:
> Could you please write unit tests for it?

Sorry, I just saw your this message now: "Just in case you are wondering: once everything has been approved, I will take out the test data that is in the file, leaving behind lines 31-150. So anything that isn't directly related to the functions will be removed before merge."

In any case, let us do our work as complete as possible, thinking as if it would exist here (instead of being migrated to MetPy anytime in the future). So, could you please initiate a unit test file for it, make a few simple but point-to-target tests, and remove your visualization tests in the script?

erogluorhan:
> > Could you please write unit tests for it?
> 
> Sorry, I just saw your this message now: "Just in case you are wondering: once everything has been approved, I will take out the test data that is in the file, leaving behind lines 31-150. So anything that isn't directly related to the functions will be removed before merge."
> 
> In any case, let us do our work as complete as possible, thinking as if it would exist here (instead of being migrated to MetPy anytime in the future). So, could you please initiate a unit test file for it, make a few simple but point-to-target tests, and remove your visualization tests in the script? (You may simply consider turning the input data and function outputs you used in your visualization test into a set of unit tests in the `test_` file)

michaelavs:
@erogluorhan I think I have added the necessary items, let me know if I missed anything!

test/test_skewt_params.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np
+from metpy.units import units
+
+from geocat.comp.skewt_params import get_skewt_vars, showalter_index
```
erogluorhan, line 4: For the code coverage to work properly, please re-write this import line as follows:

```
# Import from directory structure if coverage test, or from installed
# packages otherwise
if "--cov" in str(sys.argv):
    from src.geocat.comp import get_skewt_vars, showalter_index
else:
    from geocat.comp import get_skewt_vars, showalter_index
```

This is because, as the comment suggests, when code coverage is analyzing, it uses the directory structure, but other than code coverage, we want to test our namespace from the installed packages (to ensure this function will be callable through geocat.comp installed from conda). 

In relation to above purpose of namespace testing, our imports should be 

`from geocat.comp import ...` instead of 
`from geocat.comp.crop import ...` 

(i.e. we shouldn't specify the source script file) because we want to test if the function can be directly imported from `geocat.comp` namespace.

erogluorhan (changes requested):
We are on track. There are a single inline comment and another thing as follows:

In `/build_envs/environment.yml`, add 

`- pip`

right before

```
- pip:
      - pre-commit
```

test/test_skewt_params.py
```diff
@@ -0,0 +1,50 @@
...
+    result = showalter_index(p, tc, tdc)
+    expected = 21.353962321924012 * units.delta_degree_Celsius
+    if result == expected:
+        pass
+    else:
+        raise Exception
+
+
+def test_get_skewt_vars():
+    p_upper = np.arange(1000, 200, -50) * units.hPa
```
erogluorhan, line 34: This and other inputs below are duplicated for both unit test cases. Please migrate these from test functions to above beginning of the test script as we do in other test scripts.

test/test_skewt_params.py
```diff
@@ -0,0 +1,50 @@
...
+    p = np.append(
+        p_upper,
+        p_lower,
+    )  # Pressure levels in hPa
+    tc = np.linspace(30, -30, 23) * units.degC  # Env temp in degC
+    tdc = np.linspace(10, -30, 23) * units.degC  # DewPt temp in degC
+    pro = mpcalc.parcel_profile(p, tc[0], tdc[0])  # Parcel Profile
+
+    result = get_skewt_vars(p, tc, tdc, pro)
+    expected = 'Plcl= 747 Tlcl[C]= 6 Shox= 21 Pwat[cm]= 5 Cape[J]= 0'
```
erogluorhan, line 45: You can also move the expected result to the beginning of the test script before test cases start (for increased readability)

erogluorhan (changes requested):
Most look great! One more cycle though

erogluorhan (approved):

PR #101, opened 2021-02-05
Release prep

pilotchute:


pilotchute:
Please review this update to geocat-comp that implements f2py as the fortran backend.

meta.yaml
```diff
@@ -0,0 +1,51 @@
+package:
+  name: "geocat-comp"
+  version: "2021.01.1"
```
erogluorhan, line 3: Two points:
1. May it be better to say "2021.02.0"?
2. Looks like "version" is being used in at least two different places (meta.yml and setup.py). Instead of manually handling in multiple locations, please do not delete **version.py** and make use of it.

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,32 @@
...
+          token: ${{ secrets.A_KOOTZ_GEOCAT_COMP_TOKEN }}
+      - name: conda_env
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: temp
+          python-version: ${{ matrix.python-version }}
+          channels: conda-forge, ncar
+      - name: build_test
+        run: |
+          conda env create --file build_envs/environment_$(uname).yml
```
erogluorhan, line 29: Please move these two lines (`conda env create` and `conda activate`) to above "conda_env" step.

setup.py
```diff
@@ -1,22 +1,17 @@
-try:
-    from setuptools import setup
-    from setuptools import Extension
-except ImportError:
-    from distutils.core import setup
-    from distutils.extension import Extension
+#''' setup.py is needed, but only to make namespaces happen,
+version = '2021.01.1'
```
erogluorhan, line 2: Two points:
1. May it be better to say "2021.02.0"?
2. Looks like "version" is being used in at least two different places (meta.yml and setup.py). Instead of manually handling in multiple locations, please do not delete **version.py** and make use of it.

setup.py
```diff
@@ -1,22 +1,17 @@
...
-
-setup(name="geocat.comp",
-      package_dir={
-          '': 'src',
-          'geocat': 'src/geocat',
-          'geocat.comp': 'src/geocat/comp'
-      },
-      namespace_packages=['geocat'],
-      packages=["geocat", "geocat.comp"],
-      version=__version__,
```
erogluorhan, line 21: In regard to above version comment of mine, I think we should be still referring to version.py like here.

src/geocat/comp/version.py
```diff
@@ -1 +0,0 @@
-__version__ = "0.2"
```
erogluorhan, line 1: Again, I prefer not to remove **version.py**, but may be we can move it to root directory?

erogluorhan (changes requested):
Looks good mostly. A few points:

1) This PR is adding Github actions for build tests but not removing CircleCI bindings. Why?
2) In build environments (**build_envs** folder), there is still `geocat-ncomp` dependency but not `f2py` dependency.
3) Please address inline comments.

meta.yaml
pilotchute: setup.py now pulls version info from meta.yaml. no need for a version.py file.

src/geocat/comp/version.py
pilotchute: version now stored in meta.yaml

.github/workflows/build_test.yml
pilotchute: conda setup broken into mutiple steps, same with build test, much clearer on failures now

erogluorhan (changes requested):
In build environments (build_envs folder), please convert geocat-ncomp dependency to geocat-f2py dependency

pilotchute:
> In build environments (build_envs folder), please convert geocat-ncomp dependency to geocat-f2py dependency

Done.

erogluorhan (approved):
Great job!

pilotchute:
Will merge once the tests finish.

anissa111 (approved):
This looks great, thank you for getting this together!

PR #102, opened 2021-02-11
EOFs

erogluorhan:
Branched from "release_prep" instead of "main"

Adds `eofunc_eofs` and `eofunc_pcs`, and their unit tests.

Also adjusts some config files to reflect user and internal API changes and package dependencies.

### DIFF
--- a/test/test_eofunc.py
+++ b/test/test_eofunc.py
@@ -1,6 +1,7 @@
 from abc import ABCMeta
 from unittest import TestCase
 import numpy as np
+import numpy.testing as nt
 # from dask.array.tests.test_xarray import xr
 import xarray as xr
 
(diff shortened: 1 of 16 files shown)


erogluorhan:
The PR seems to be failing for CircleCI tests but I think we can ignore them since Github actions are all passing. I did not change CircleCI tests because they are heavily based on `geocat-ncomp` plus `libncomp` stack, and we will likely get rid of CircleCI tests.

test/test_eofunc.py
```diff
@@ -0,0 +1,324 @@
...
+
+        results = eofunc_eofs(data, neofs=1, time_dim=2)
+        eof = results.data
+        attrs = results.attrs
+
+        self.assertEqual((1, 4, 4), eof.shape)
+
+        for e in np.nditer(eof):
+            self.assertAlmostEqual(0.25, e, 2)
+
```
andersy005, line 135: You could use the [Numpy's testing routines](https://numpy.org/doc/stable/reference/routines.testing.html) here instead of having to loop over the array.

erogluorhan: Yeah, thanks for catching! Addressed that now.

FYI: We also use  Numpy's testing routines in our tests; however, we had to inherit recently a lot of older codes from geocat-ncomp stack to deprecate it, and we may have missed some of those older codes such as here. First-priority was having the functionality in geocat-f2py and not to miss any function from soon-to-be-deprecated geocat-ncomp.

Though, your reviews are highly helpful and appreciated!

pilotchute (approved):
Looks great! Thanks Orhan!

anissa111 (approved):
Assuming all the non CI tests pass this looks good!

PR #111, opened 2021-02-24
Vinth2p (interp_hybrid_to_pressure)

erogluorhan:
This PR adds the `interp_hybrid_to_pressure` that implements NCL's `vinth2p`.

- It adds some dependencies (for both implementation and testing) in several places


src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,111 @@
...
+                              p0=100000.,
+                              new_levels=__pres_lev_mandatory__,
+                              lev_dim='lev',
+                              method='linear'):
+    """
+    Interpolate data from hybrid-sigma levels to isobaric levels.
+
+    ACKNOWLEDGEMENT: We'd like to thank to Brian Medeiros (https://github.com/brianpm), Matthew Long
+    (https://github.com/matt-long), and Deepak Cherian (https://github.com/dcherian) at NCAR for their
+    great contributions since the code implemented here is mostly based on their work.
```
andersy005, line 33: I'd add this to a [`Notes`](https://numpydoc.readthedocs.io/en/latest/format.html#sections) section right after `Parameters`. 

```suggestion
    Notes
    -----
    ACKNOWLEDGEMENT: We'd like to thank to Brian Medeiros (https://github.com/brianpm), Matthew Long
    (https://github.com/matt-long), and Deepak Cherian (https://github.com/dcherian) at NCAR for their
    great contributions since the code implemented here is mostly based on their work.
```

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Parameters
+    ----------
+    data : `xarray.DataArray`:
```
andersy005, line 37: Per [numpydoc style](https://numpydoc.readthedocs.io/en/latest/format.html), the back-ticks aren't necessary. 

For instance: 

```suggestion
    data : xarray.DataArray
```

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,111 @@
...
+        dimension size as the `lev_dim` dimension of data.
+
+    p0 :
+        Scalar numeric value equal to surface reference pressure (Pa).
+
+    new_levels : the output pressure levels (Pa)
+        A one-dimensional array of output pressure levels (Pa). If not given, the mandatory
+        list of 21 pressure levels is used.
+
+    lev_dim : `string`:
```
andersy005, line 54:
```suggestion
    lev_dim : str
```

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,111 @@
...
+    return hya * p0 + hyb * psfc
+
+
+def interp_hybrid_to_pressure(data,
+                              ps,
+                              hyam,
+                              hybm,
+                              p0=100000.,
+                              new_levels=__pres_lev_mandatory__,
+                              lev_dim='lev',
```
andersy005, line 26: Can we set this to None by default instead of 'lev', and then use `cf_xarray` to retrieve the name of the vertical dim name? 

I imagine something along these lines:

```python
if lev_dim is None:
  try:
     data.cf["vertical"].name 
   except Exception:
      raise ValueError("Unable to determine vertical dimension name. Please specify the name via `lev_dim` argument.')
```

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,111 @@
...
+
+        return func_interpolate(plev, pressure, data, axis=interp_axis)
+
+    # Apply vertical interpolation
+    # Apply Dask parallelization with xarray.apply_ufunc
+    output = xr.apply_ufunc(
+        _vertical_remap,
+        new_levels,
+        pressure.values,
+        data.values,
```
andersy005, line 85: Pass input data as is. `.values` eagerly loads a dask array into a numpy array, and this is not ideal in most cases:


```suggestion
        data,
```

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,111 @@
...
+           ] + [data.dims[i] for i in range(data.ndim) if i != interp_axis]
+
+    coords = {}
+    for (k, v) in data.coords.items():
+        if k != lev_dim:
+            coords.update({k: v})
+        else:
+            coords.update({"plev": new_levels})
+
+    output = xr.DataArray(output, dims=dims, coords=coords)
```
andersy005, line 109: Since `xr.apply_ufunc` returns an xarray object, I don't think we need to instantiate an xarray.DataArray object again. We can update the coords and dims without an xarray.DataArray instantiation.

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,111 @@
...
+
+    interp_axis = data.dims.index(lev_dim)
+
+    # Define interpolation function
+    if method == 'linear':
+        func_interpolate = metpy.interpolate.interpolate_1d
+    elif method == 'log':
+        func_interpolate = metpy.interpolate.log_interpolate_1d
+    else:
+        raise ValueError(f'Unknown interpolation method: {method}')
```
andersy005, line 72:
```suggestion
        raise ValueError(f'Unknown interpolation method: {method}. Supported methods are: "log" and "linear".')
```

src/geocat/comp/interp_hybrid_to_pressure.py
andersy005: If you decide to go this route, you will need to import cf_xarray in this module.

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,115 @@
+import metpy.interpolate
+import numpy as np
+import xarray as xr
+
+__pres_lev_mandatory__ = np.array([
+    1000, 925, 850, 700, 500, 400, 300, 250, 200, 150, 100, 70, 50, 30, 20, 10,
+    7, 5, 3, 2, 1
+])  # Mandatory pressure levels (mb)
+__pres_lev_mandatory__ = __pres_lev_mandatory__ * 100.0  # Convert mb to Pa
+
```
andersy005, line 11: The default dtype for this ends up being `float64` which can cause unnecessary memory spikes in some scenarios. For instance, if the user provides `float32` data arrays as inputs, we end up doing the computation in double precision (which I think is unnecessary). I'm wondering if we should cast this to an appropriate default dtype (e.g. `float32`) so as to let user control the level of precision (single, double) to use?

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Done, thanks for suggesting this. (Will also address this in other scripts soon)

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Addressed

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Done. Thanks for cf_xarray suggestion!

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Actually attempted to do so but could not figure out the reason(s) of dimensionality exceptions so far. Converted `.values` to `.data` for now, thinking that it can allow to use DaskArray as is without a conversion to numpy.ndarray.

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Related to the above `.data` thing, apply_ufunc returns ndarray, so I need to instantiate a xarray.dataarray for now

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Done, thanks!

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Agreed & done, and thanks for catching this!

erogluorhan:
@andersy005 I don't know why but I cannot re-request your reviews (It happened once again before). Anyway, I addressed your reviews; please feel free to give it another review.

erogluorhan:
Hi @anissa111 @pilotchute and @andersy005 , sorry for pushing for this, but to release the new GeoCAT-comp version (v2021.03.0) in a few days, could you please give this a review at your best convenience?

pilotchute (approved):

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,124 @@
...
+        pressure.data,
+        new_levels,
+        interp_axis,
+        exclude_dims=set((lev_dim,)),  # Set dimensions allowed to change size
+        input_core_dims=[[lev_dim], [lev_dim], ["plev"], []], # Set core dimensions
+        output_core_dims=[["plev"]],  # Specify output dimensions
+        vectorize=True,  # loop over non-core dims
+        dask="parallelized",  # Dask parallelization
+        output_dtypes=[data.dtype],
+    )
```
andersy005, line 109: @erogluorhan, this code doesn't want when the input when `data` is an xarray object backed by dask.array. for instance, if you chunk `_ps` in your test, you will notice that you get an error: 

`ValueError: cannot use dask='parallelized' for apply_ufunc unless at least one input is an xarray object`

```python
In [20]: _ps = _ps.chunk()

In [21]: _ps
Out[21]: 
<xarray.DataArray 'PS' (time: 1, lat: 64, lon: 128)>
dask.array<xarray-<this-array>, shape=(1, 64, 128), dtype=float32, chunksize=(1, 64, 128), chunktype=numpy.ndarray>
Coordinates:
  * lat      (lat) float32 -87.86 -85.1 -82.31 -79.53 ... 79.53 82.31 85.1 87.86
  * lon      (lon) float32 0.0 2.812 5.625 8.438 ... 348.8 351.6 354.4 357.2
  * time     (time) float64 791.0
Attributes:
    long_name:  surface pressure
    units:      Pa
    time_op:    average

In [20]: _ps = _ps.chunk()

In [21]: _ps
Out[21]: 
<xarray.DataArray 'PS' (time: 1, lat: 64, lon: 128)>
dask.array<xarray-<this-array>, shape=(1, 64, 128), dtype=float32, chunksize=(1, 64, 128), chunktype=numpy.ndarray>
Coordinates:
  * lat      (lat) float32 -87.86 -85.1 -82.31 -79.53 ... 79.53 82.31 85.1 87.86
  * lon      (lon) float32 0.0 2.812 5.625 8.438 ... 348.8 351.6 354.4 357.2
  * time     (time) float64 791.0
Attributes:
    long_name:  surface pressure
    units:      Pa
    time_op:    average

In [22]: u_int = geocat.comp.interp_hybrid_to_pressure(_data,
    ...:                                                       _ps[0, :, :],
    ...:                                                       _hyam,
    ...:                                                       _hybm,
    ...:                                                       p0=_p0,
    ...:                                                       new_levels=_pres3d,
    ...:                                                       method="log")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-22-86cfb5bc43ab> in <module>
      5                                                       p0=_p0,
      6                                                       new_levels=_pres3d,
----> 7                                                       method="log")

~/devel/ncar/geocat-comp/src/geocat/comp/interp_hybrid_to_pressure.py in interp_hybrid_to_pressure(data, ps, hyam, hybm, p0, new_levels, lev_dim, method)
    106         vectorize=True,  # loop over non-core dims
    107         dask="parallelized",  # Dask parallelization
--> 108         output_dtypes=[data.dtype],
    109     )
    110 

~/opt/miniconda/envs/geocat_comp_build/lib/python3.7/site-packages/xarray/core/computation.py in apply_ufunc(func, input_core_dims, output_core_dims, exclude_dims, vectorize, join, dataset_join, dataset_fill_value, keep_attrs, kwargs, dask, output_dtypes, output_sizes, meta, dask_gufunc_kwargs, *args)
   1139     else:
   1140         # feed anything else through apply_array_ufunc
-> 1141         return apply_array_ufunc(func, *args, dask=dask)
   1142 
   1143 

~/opt/miniconda/envs/geocat_comp_build/lib/python3.7/site-packages/xarray/core/computation.py in apply_array_ufunc(func, dask, *args)
    782         elif dask == "parallelized":
    783             raise ValueError(
--> 784                 "cannot use dask='parallelized' for apply_ufunc "
    785                 "unless at least one input is an xarray object"
    786             )

ValueError: cannot use dask='parallelized' for apply_ufunc unless at least one input is an xarray object

```

One remedy is to pass `data` instead of `data.data` as input to `xr.apply_ufunc()` ( I haven't test this though).

andersy005 (changes requested):

src/geocat/comp/interp_hybrid_to_pressure.py
andersy005: It would be useful to add tests that makes sure that the function works for xarray objects backed by dask.array.

erogluorhan: Thanks for catching this, Anderson! I will address this soon.

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,122 @@
...
+    def _vertical_remap(data, pressure):
+        """
+        Define interpolation function.
+        """
+
+        return func_interpolate(new_levels, pressure, data, axis=interp_axis)
+
+    # Apply vertical interpolation
+    # Apply Dask parallelization with xarray.apply_ufunc
+    output = xr.apply_ufunc(
```
erogluorhan, line 97: Hey @dcherian , I am having an issue with `apply_ufunc` and wondering if you would have some time to look at it and a quick solution to it:

I think that I set up a correct use of`xarray.DataArray.apply_ufunc` for my `_vertical_remap ` function on two data arrays of the same dimensions ['lev', 'lat', 'lon'] and the output array of ['plev', 'lat', 'lon'] where `plev` is the new (i.e. changed) `lev` dimension coming from `new_levels`. i.e. `lat` and `lon` stays the same, naturally in this 1d interpolation.

When my input data are **unchunked** `xarray.DataArray`s, function runs successfully (e.g. `test_interp_hybrid_to_pressure_atmos` and `test_interp_hybrid_to_pressure_atmos_plot` in `test/test_interp_hybrid_to_pressure.py`). However, when I make the input data to be **chunked** `xarray.DataArray` (e.g. `test_interp_hybrid_to_pressure_atmos_dask` in `test/test_interp_hybrid_to_pressure.py`), I get the following error:

`ValueError: dimension 'plev' in 'output_core_dims' needs corresponding (dim, size) in 'output_sizes'`

which I could not figure out so far (Though I am walking through [apply_ufunc on unvectorized data doc](https://xarray.pydata.org/en/stable/examples/apply_ufunc_vectorize_1d.html) ).

Would you have an idea on this?

dcherian: As it says, you need to tell it the size of `plev` in the output.

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -0,0 +1,122 @@
...
+        _vertical_remap,
+        data,
+        pressure,
+        exclude_dims=set((lev_dim,)),  # Set dimensions allowed to change size
+        input_core_dims=[[lev_dim], [lev_dim]],  # Set core dimensions
+        output_core_dims=[["plev"]],  # Specify output dimensions
+        vectorize=True,  # loop over non-core dims
+        dask="parallelized",  # Dask parallelization
+        output_dtypes=[data.dtype],
```
dcherian, line 106:
```suggestion
        dask_gufunc_kwargs={"output_sizes": {"plev": len(new_levels)}},
```

like this?

erogluorhan: Ah yes, thanks so much!  I was stuck thinking on and playing with `output_core_dims`.

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Addressed this now

andersy005 (approved):
Looks good to me 👍

PR #112, opened 2021-02-25
Conda builds from release tags and changed version settings

### LINKED ISSUE #109: Tagging Releases
Today, it's not possible to checkout a specific release version (tag) of geocat-comp via git. I'm wondering if there's any plan for tagging future releases in this repository? I wanted to compare changes between two releases via Git. Having access to git tags/releases would make this easy.

erogluorhan:
Closes #109 

In an effort to respond to above issue, this PR modifies `meta.yml` and other relevant files to fetch release version from this repo's tags instead of source code in `main` branch.

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -1,6 +1,8 @@
+{% set version = "2021.02.0" %}
+
 package:
   name: 'geocat-comp'
-  version: '2021.02.0'
+  version: {{ version }}
 
 build:
   noarch: python
@@ -8,8 +10,9 @@ build:
   skip: True  # [py<3.5]
 
 source:
-  git_rev: main
-  git_url: https://github.com/NCAR/geocat-comp.git
+#  git_rev: main
+#  git_url: https://github.com/NCAR/geocat-comp.git
+  url: https://github.com/NCAR/geocat-comp/archive/v{{ version }}.tar.gz   # Fetch from Github repo tags
 
 requirements:
   host:
--- a/meta.yaml
+++ b/meta.yaml
@@ -10,9 +10,9 @@ build:
   skip: True  # [py<3.5]
 
 source:
-#  git_rev: main
-#  git_url: https://github.com/NCAR/geocat-comp.git
-  url: https://github.com/NCAR/geocat-comp/archive/v{{ version }}.tar.gz   # Fetch from Github repo tags
+  git_rev: v{{ version }}
+  git_url: https://github.com/NCAR/geocat-comp.git
+  #url: https://github.com/NCAR/geocat-comp/archive/v{{ version }}.tar.gz   # Fetch from Github repo tags
 
 requirements:
   host:
--- a/meta.yaml
+++ b/meta.yaml
@@ -10,9 +10,10 @@ build:
   skip: True  # [py<3.5]
 
 source:
-  git_rev: v{{ version }}
+  #git_rev: main   # For debugging purposes
+  git_rev: v{{ version }}   # For pulling tag as a branch
   git_url: https://github.com/NCAR/geocat-comp.git
-  #url: https://github.com/NCAR/geocat-comp/archive/v{{ version }}.tar.gz   # Fetch from Github repo tags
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -2,14 +2,12 @@
 from setuptools import setup, find_packages
 from pathlib import Path
 
-
 #''' moved into function, can now be used other places
 def version():
```
andersy005, line 6: Wouldn't it be more robust to use git as the source of truth for the version instead of the string in the YAML file? https://pypi.org/project/setuptools-scm/ is a good candidate for this. 

Also, with this current approach, it's not possible to retrieve the version information at runtime. It'd be useful to provide a convenient way to retrieve version. For instance:

```python

In [3]: import geocat.comp

In [4]: import xarray as xr

In [5]: import numpy as np

In [6]: xr.__version__
Out[6]: '0.16.2'

In [7]: np.__version__
Out[7]: '1.19.5'

In [8]: geocat.comp.__version__
------------------------------------------------------------------------
AttributeError                         Traceback (most recent call last)
<ipython-input-8-a97600b4156d> in <module>
----> 1 geocat.comp.__version__

AttributeError: module 'geocat.comp' has no attribute '__version__'
```

pilotchute: Anderson, our current approach is a good compromise that meets our needs for lack of duplication and ease of use. While it would be nice to have geocat.comp.__version__ store the build version, it doesn't seem to be strictly needed. (also since this is a conda package, `conda list` would be more than sufficient for getting version information)

meta.yaml
```diff
@@ -1,15 +1,18 @@
+{% set version = "2021.02.0" %}
+
 package:
   name: 'geocat-comp'
-  version: '2021.02.0'
+  version: {{ version }}
```
pilotchute, line 5: Why this change? What does it do that improves on the way the version was being set?

setup.py
erogluorhan: I am migrating @andersy005 's suggestion into a [new issue](https://github.com/NCAR/geocat-comp/issues/114)  as a fun activity for myself and whenever I have some free time, I'd be happy to play with it.

meta.yaml
erogluorhan: It is related to this (Line 15 in the same file):

`url: https://github.com/NCAR/geocat-comp/archive/v{{ version }}.tar.gz   # Fetch from Github repo tags`

We are slightly changing our `meta.yml` to use Github tag instead of source in the `main' branch. However, for reading "version" in this line, I needed this small change of initiation for "version". Nothing functional.

setup.py
andersy005: > While it would be nice to have geocat.comp.version store the build version, it doesn't seem to be strictly needed.

If we assume that everyone is using the stable release, I agree with this statement. However, what happens when I install geocat-comp from source via 

```bash
python -m pip install git+https://github.com/NCAR/geocat-comp.git 
```
???


This works as of today. However, the version information you get from `conda list` are misleading. I do think that in this case (for install from source) providing version information at runtime is very useful. As a user, I find it necessary to be able to know which git commit I am on. For instance, the following tells me that there has been 342 commits to master since the `v0.16.1` release, and I happen to be on git commit with sha: `0cff30e1`


```python
In [1]: import xarray as xr

In [2]: xr.__version__
Out[2]: '0.16.1.dev342+g0cff30e1.d20210220'
```

I find knowing this kind of information to be valuable since you can pin point source of problems during development, etc... These are my 2¢ as a user who installs packages from source, and if I am the only geocat-comp user who finds this information to be useful, I am happy to find a workaround :)

erogluorhan: Thanks a lot for the detailed explanation, @andersy005 ! I find it valuable and, again, am happy to work on this at my first convenience.

erogluorhan:
Hi @anissa111 @pilotchute and @andersy005 , sorry for pushing for this, but to release the new GeoCAT-comp version (v2021.03.0) in a few days, could you please give this a review at your best convenience?

pilotchute (approved):

Issue #57, opened 2020-03-05
Distributing geocat via PyPI/pip


andersy005:
As it's been pointed out in this comment on [Discus](http://disq.us/p/27q4m1e), **the only way to use geocat today is via conda**. This may end up penalizing/discouraging non-conda users or may force them to migrate to conda --which is not convenient for everyone-- if they really want to use geocat. It would be useful to consider building and publishing **wheels** to PyPI. 

@khallock pointed out:

> In the past, we provided binary releases of our NCL-related software, but doing so means that we have to maintain a large collection of build environments (particularly for different flavors of Linux)

However, it's my understanding that this process can be less time consuming when done via automation. In the past, I've used GitHub actions in conjunction with [cibuildwheel](https://github.com/joerick/cibuildwheel) to build wheels for windows, MacOS, Linux, and to publish them to PyPI. 

Let me know if this is of interest, and I will open a PR.

Cc @matt-long, @kmpaul

P.S.: I wasn't sure whether this is the right repo to raise this issue in. Let me know if there's an appropriate place for this discussion.

clyne:
Good points, @andersy005. Thanks for bringing this up. The GeoCAT will definitely discuss f2f.

andersy005:
Now that `geocat-f2py` is an optional dependency,  It's my understanding that publishing `geocat-comp` to PyPI is now feasible, and should be straightforward.

PR #115, opened 2021-02-26
GitHub action for publishing geocat-comp to PyPI

andersy005:
Now that #109 has been addressed, it is possible to publish geocat-comp to PyPI. This GitHub action workflow is meant to be triggered whenever a new GitHub release is published. 

Towards fixing #57

@NCAR/geocat, for this to properly work, someone needs to 

1. Create an account on pypi.org
2. Create an API token from https://pypi.org/manage/account/ that will be used to upload the built package. 
3. Add the API token as a secret to this GitHub repo settings. The secret's name must be set to `PYPI_PASSWORD` so as to match what is declared in `pypi.yaml` file


erogluorhan (commented):
It seems to be failing for Python 3.9 for both Ubuntu and MacOS for different steps

setup.py
```diff
@@ -10,15 +28,31 @@ def version():
...
-
 setup(
     name='geocat.comp',
     version=version(),
+    maintainer='geocat team',
+    maintainer_email='geocat@ucar.edu',
+    python_requires='>=3.6',
+    install_requires=requirements,
+    description=
+    'GeoCAT-comp is a Python wrapper around a Fortran library containing a subset of NCL (NCAR Command Language) computational routines.',
```
erogluorhan, line 39: I would not prefer mentioning Fortran for geocat-comp release. May be some thing as follows or close:

"GeoCAT-comp is computational component of the GeoCAT project and provides implementations of computational functions for analysis of geosciences data"

setup.py
```diff
@@ -10,15 +28,31 @@ def version():
         if index > -1:
             return line[index + 15:].replace('\" %}', '').strip()
 
-
 setup(
     name='geocat.comp',
     version=version(),
+    maintainer='geocat team',
+    maintainer_email='geocat@ucar.edu',
```
erogluorhan, line 35: Do you know for what purposes and with what frequency this email would be used by PyPI? Asking because geocat@ucar.edu is a mailing list rather than being an individual mail account managed by someone.

setup.py
```diff
@@ -2,6 +2,24 @@
 from setuptools import setup, find_packages
 from pathlib import Path
 
+CLASSIFIERS = [
```
erogluorhan, line 5: What about rolling this block (Lines 5 - 21) up with an informative begin and end comments that this is for PyPI?

erogluorhan (commented):
It's a great contribution overall, thanks! Please see a few inline comments.

setup.py
andersy005: PyPI doesn't use this email, and this entry is optional. I think the purpose is for users to know how to reach out to the package publisher via email if need be.

setup.py
andersy005: Sounds good!

setup.py
andersy005: For less confusion, I just moved those lines into `setup()` directly.

erogluorhan (approved):
Looks great! 

I have added PYPI_PASSWORD secret to the repo.

erogluorhan:
@pilotchute Could you please review this PR when you schedule permits?

pilotchute:
Looking it over now

pilotchute (approved):
I'm apprehensive about the changes to setup.py, but I'm approving this since I see no functional issues with the changes.

PR #119, opened 2021-03-03
Fix codecov

erogluorhan:
The current "Upload code coverage to Codecov" in `ci.yml` cannot generate coverage reports (due to this warning `Coverage.py warning: No data was collected. (no-data-collected)`, see in [Latest main CI workflows](https://github.com/NCAR/geocat-comp/runs/2023837876?check_suite_focus=true)). As a result, codecov badge in README is not working.

This PR simply fixes it (e.g. See [CI workflow of this branch](https://github.com/NCAR/geocat-comp/runs/2023837876?check_suite_focus=true))


.github/workflows/ci.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       
       - name: Running Tests
         run: |
-          python -m pytest test -v --cov=./src --cov-report=xml 
+          python -m pytest test -v --cov=./ --cov-report=xml
```
andersy005, line 49: @erogluorhan, this is reporting the wrong coverage:

Screen Shot 2021-03-03 at 1 11 41 PM

The coverage from `test/*.py` isn't really useful. We actually care about the source code in `src/geocat/comp/*.py`. 

Try the following and see if it makes a difference:

```suggestion
          python -m pytest test -v --cov=./src/geocat/comp --cov-report=xml
```

erogluorhan: Thank you for letting me know this! The new code now seems to be working, at least generating the coverage report for `src/geocat/comp`. However, coverage seems to be **0%**, which shouldn't be the case in existence of our unit tests under `/test` folder. Apparently I need to increase my understanding on how this report is being generated.

erogluorhan:
@anissa111 @pilotchute Could you please review this PR when you schedule permits?

test/test_climatology.py
```diff
@@ -2,7 +2,7 @@
 import xarray as xr
 import numpy as np
 import pandas as pd
-import geocat.comp
+from src.geocat.comp import anomaly, climatology, month_to_season
```
andersy005, line 5: @erogluorhan, I find this line to be less intuitive (especially for folks not familiar with the structure of the repository). I'm curious as to why the directory nesting is necessary. 


```bash
src
├── geocat
│   └── comp
│       ├── __init__.py
│       ├── __pycache__
│       ├── climatology.py
│       ├── dewtemp.py
│       ├── eofunc.py
│       ├── errors.py
│       ├── interp_hybrid_to_pressure.py
│       ├── polynomial.py
│       └── relhum.py
```


Could we just remove the `src` and move the `geocat` at the root directory i.e: `geocat-comp/geocat`, and then keep the original import `geocat.comp import anomaly, climatology, month_to_season`:


```bash
geocat
└── comp
    ├── __init__.py
    ├── __pycache__
    ├── climatology.py
    ├── dewtemp.py
    ├── eofunc.py
    ├── errors.py
    ├── interp_hybrid_to_pressure.py
    ├── polynomial.py
    └── relhum.py
```

erogluorhan: Removing `/src` and moving the `geocat/comp` at the root directory makes sense to me, but wouldn't there be an amiguity in that case as follows:

In test scripts or any build-from-code activity, when you do your imports as `from geocat.comp import foo` , would it go find your function from your source directory or from the installed `geocat.comp` package namespace under your active conda environment (You might have installed `geocat.comp` via conda or via your own source with `pip install .`)?

In my opinion, any test script should be testing the code against its installed package within the active conda environment. If it is so, then even my `src.geocat.comp` is not correct, and we should instead find a way for `codecov` to report correct coverage through namespace imports (instead of source directory parsing).

I hope I am clear to describe my thoughts.

Any thoughts? cc'ing @pilotchute

anissa111: I was under the impression that the nested directory structure was due to namespace concerns, too, but if the structure could be simplified, I would be all for it.

andersy005: As I understand it, the namespaces are configured here: 

https://github.com/NCAR/geocat-comp/blob/ede232f93ad8fca4b10de16ff74852587027c9fa/setup.py#L17-L21


and this could be easily updated to:

```python
 package_dir={ 
     'geocat': 'geocat', 
     'geocat.comp': 'geocat/comp' 
 }
```

erogluorhan: > As I understand it, the namespaces are configured here:
> 
> https://github.com/NCAR/geocat-comp/blob/ede232f93ad8fca4b10de16ff74852587027c9fa/setup.py#L17-L21
> 
> and this could be easily updated to:
> 
> ```python
>  package_dir={ 
>      'geocat': 'geocat', 
>      'geocat.comp': 'geocat/comp' 
>  }
> ```

Yes namespaces can be configured like this, but that's not what I meant. Let me be more precise:

Let's assume we did this namespace change and we did `pip install .` (or `conda install geocat-comp`).  Assume  we have the following sample import in our test script:

`from geocat.comp import interp_hybrid_to_pressure`

My question was that which one of the followings would this function be fetched from in above import statement:

1. `$GEOCATCOMPROOT/geocat/comp/interp_hybrid_to_pressure.py`
2. `miniconda3/envs/<env_name>/lib/python3.x/site-packages/geocat/comp/interp_hybrid_to_pressure.py`

Furthermore, I was wondering if there would be a confusion / difference of how this import is handled from e.g. PyCharm to bash, or from one OS to other. 

I can give this a try in my local but wanted to discuss here in broad. 

Note: My personal view is that we should both have our test scripts and `codecov` configured in such a way that we always guarantee imports from option 2 above (i.e. conda package). Therefore, I now believe my so-called correction of `from src.geocat.comp import function` is not correct indeed.

andersy005: > from geocat.comp import interp_hybrid_to_pressure

> My question was that which one of the followings would this function be fetched from in above import statement:

> 1. $GEOCATCOMPROOT/geocat/comp/interp_hybrid_to_pressure.py
> 2. miniconda3/envs/<env_name>/lib/python3.x/site-packages/geocat/comp/interp_hybrid_to_pressure.py

- The directory from which the script is run takes precedence over any other options (for instance  list of directories contained in the PYTHONPATH environment variable or other installation-dependent list of directories configured at the time Python is installed). So, if your run the import inside the `$GEOCATCOMPROOT/`, you will get option 1. Run the import from other location, would result in option 2 with one exception (See the next point)

- If you install the package in editable mode i.e. `python -m pip install -e .`, Python creates a symlink to the modules in `$GEOCATCOMPROOT/`. With this option, you'd get option 1 regardless of the directory from which the script is invoked from.

andersy005 (approved):
Thank you for working on this, @erogluorhan!

test/test_climatology.py
erogluorhan: > > from geocat.comp import interp_hybrid_to_pressure
> 
> > My question was that which one of the followings would this function be fetched from in above import statement:
> 
> > 1. $GEOCATCOMPROOT/geocat/comp/interp_hybrid_to_pressure.py
> > 2. miniconda3/envs/<env_name>/lib/python3.x/site-packages/geocat/comp/interp_hybrid_to_pressure.py
> 
> * The directory from which the script is run takes precedence over any other options (for instance  list of directories contained in the PYTHONPATH environment variable or other installation-dependent list of directories configured at the time Python is installed). So, if your run the import inside the `$GEOCATCOMPROOT/`, you will get option 1. Running the import from other locations, would result in option 2 with one exception (See the next point)
> * If you install the package in editable mode i.e. `python -m pip install -e .`, Python creates a symlink to the modules in `$GEOCATCOMPROOT/`. With this option, you'd get option 1 regardless of the directory from which the script is invoked from.

Thanks for the clarification!

I just have incorporated `src/geocat/comp` to `geocat/comp` change, and we seem to have everything working fine (codecov, setup.py, conda build, etc.). Feel free to give it another review.

erogluorhan:
> Thank you for working on this, @erogluorhan!

Thanks a lot for your discussion!

Btw, accidentally re-requested a review after your approval.

andersy005 (approved):

pilotchute (changes requested):
The /src directory shouldn't be removed. we should keep the current directory structure.

erogluorhan:
> The /src directory shouldn't be removed. we should keep the current directory structure.

Is there a specific reason for this that I am missing? (I tested the namespace and imports after this change and nothing seemed wrong or broken to me. )

pilotchute:
> > The /src directory shouldn't be removed. we should keep the current directory structure.
> 
> Is there a specific reason for this that I am missing? (I tested the namespace and imports after this change and nothing seemed wrong or broken to me. )

Because it makes it much harder to see namespace issues. 

With the `./src/geocat/comp/funcname.py` structure, for anything being run in the repository's base directory (pytest for example), it is very clear which code is being run/imported, the package that was installed via `pip install` or the uninstalled code in `./src/geocat/...`

If `./src/geocat` is moved to `./geocat`, then anything run in the repository's base directory, doesn't distinguish between the installed package, or the uninstalled code. 

I strongly oppose this particular change.

erogluorhan:
> I will start to work on this branch to allow codecov to work using the ./src directory structure.

Sounds good, thanks!

erogluorhan:
> I will start to work on this branch to allow codecov to work using the ./src directory structure.

For your convenience, I reverted all of my `/src` removal changes and pushed now. Pull before starting to work.

pilotchute (approved):

PR #129, opened 2021-03-18
A better fix for codecov

pilotchute:
using paired tests. 


pilotchute:
Granted while I'm the implementer of this fix, I do not like that it duplicates our tests.

erogluorhan (commented):
I am happy to see this is a solution to keep our directory structure as is and have `codecov` work correctly; however, duplicating the test code by pairing every single test file for `codecov` does not sound like the best option to me (Though we couldn't find a workaround that fits best to our practices yet).

I am curious if such a workaround would be doable: in a single test file as we did before, writing the import statement depending on if `pytest` or `pytest-cov` is being called in that specific test run. Which might look similar to following:

```
if pytest-cov:
   from src.geocat.comp import functionname
elif pytest:
   from geocat.comp import functionname
```    

Thoughts?

pilotchute:
Worth trying:
```
if "pytest-cov" in sys.modules:
  import src.geocat.comp ....
else :
  import geocat.comp....
```

pilotchute:
The if/else branching doesn't work, but I've found some strange behavior we can take advantage of.

pilotchute:
Ok, so, codecov basically just tracks what is loaded into memory during the test, and compares what was loaded against what is in the directory that it's supposed to measure coverage of.

pilotchute:
Note that codecov does not consider anything used from an installed package as "loaded into memory".

pilotchute:
So, we can have a python file that imports whatever we want from src.geocat, and even if that file does nothing other than import the code and then exit (removing those imports from memory again). codecov will sill see it as being "covered by testing".

I do not need to say how useless of a measurement this is. 
A file containing only the line: `from src.geocat.comp import *` gets the following codecov report:

```
---------- coverage: platform darwin, python 3.9.2-final-0 -----------
Name                                           Stmts   Miss  Cover
------------------------------------------------------------------
src/geocat/comp/__init__.py                       11      2    82%
src/geocat/comp/climatology.py                    62     48    23%
src/geocat/comp/dewtemp.py                        23     19    17%
src/geocat/comp/eofunc.py                         65     55    15%
src/geocat/comp/errors.py                         12      0   100%
src/geocat/comp/interp_hybrid_to_pressure.py      32     24    25%
src/geocat/comp/polynomial.py                    189    173     8%
src/geocat/comp/relhum.py                         84     76    10%
------------------------------------------------------------------
TOTAL                                            478    397    17%
```

erogluorhan:
@pilotchute I now have read all of your investigations above, and thanks a lot for digging into this!

I agree with your observations about its being analyzing the assets that are being loaded into memory only. However, I managed to solve the import-branching problem as we discussed above. Below is a sample import block that understands if we are running a regular build test or a test for code coverage:

`
import sys
if "--cov" in str(sys.argv):
    from src.geocat.comp import function(s)
else:
    from geocat.comp import functions(s)
`

I gave it a try and saw it is working. I also dug into the the report `codecov` generated for every single code script. It seems like it is able to catch the code blocks we are covering with our test cases and those we couldn't cover (I observed these on `interp_hybrid_to_pressure`).

Therefore, I removed duplicate test files, incorporated the above fix to all of our original test files, and pushed them here. Sorry for pushing to your PR but I wanted you to see them in place. If we find any concerns, I am happy to revert them all.

erogluorhan:
> @pilotchute I now have read all of your investigations above, and thanks a lot for digging into this!
> 
> I agree with your observations about its being analyzing the assets that are being loaded into memory only. However, I managed to solve the import-branching problem as we discussed above. Below is a sample import block that understands if we are running a regular build test or a test for code coverage:
> 
> ```
> import sys
> if "--cov" in str(sys.argv):
>     from src.geocat.comp import function(s)
> else:
>     from geocat.comp import functions(s)
> ```
> 
> I gave it a try and saw it is working. I also dug into the the report `codecov` generated for every single code script. It seems like it is able to catch the code blocks we are covering with our test cases and those we couldn't cover (I observed these on `interp_hybrid_to_pressure`, [see it here](https://codecov.io/gh/NCAR/geocat-comp/src/9c52bbede6949958367dc2eb4dac0dfbaf01a5ed/src/geocat/comp/interp_hybrid_to_pressure.py)).
> 
> Therefore, I removed duplicate test files, incorporated the above fix to all of our original test files, and pushed them here. Sorry for pushing to your PR but I wanted you to see them in place. If we find any concerns, I am happy to revert them all.

As a follow up on this:

If we decide this current shape works for us, we should do two additional things as follows before merging the PR (I can do both of them):

1. Remove the commented-out, single-line import lines from 6 test files
2. Since we are now checking for "--cov" in our `pytest` runs, we should split the current "Run tests" into two new jobs "Run namespace tests" and "Run coverage tests" in `ci.yml, maybe as follows:

```
- name: Run Namespace Tests
        run: |
          python -m pytest test

- name: Run Coverage Tests
        run: |
          python -m pytest test -v --cov=./src/geocat/comp --cov-report=xml
```

pilotchute (commented):
Orhan since I started this pull I can't be the approver, that said, I approve.

erogluorhan (approved):
Thanks for looking into this and exploring it!

erogluorhan:
> Orhan since I started this pull I can't be the approver, that said, I approve.

Thanks Austin! I was waiting for tests to succeed before approving it.

PR #138, opened 2021-04-07
heat_index

anissa111:
This is a work in progress PR for the implementation of `heat_index_nws`, renamed here `heat_index`, as described in [S@S Scenario 2](https://github.com/NCAR/science-at-scale/issues/2).


anissa111:
Just a note, during the testing process for heat_index, @michaelavs and I discovered a bug in the original ncl. Following the original [NWS equations](https://www.wpc.ncep.noaa.gov/html/heatindex_equation.shtml), one of the "adjustments" was not being calculated in the ncl function. The values produced with this python version were verified by hand according to the original equations where the bug was present. This is why the ncl test script and ncl ground truth file were removed from this function's testing.

src/geocat/comp/heat_index.py
```diff
@@ -0,0 +1,150 @@
...
+    if rh < 13 and (80 < t < 112):
+        hi = hi - ((13 - rh) / 4) * np.sqrt((17 - abs(t - 95)) / 17)
+
+    if rh > 85 and (80 < t < 87):
+        hi = hi + ((rh - 85.0) / 10.0) * ((87.0 - t) / 5.0)
+
+    return hi
+
+
+def heat_index(t, rh, alt_coef=False):
```
erogluorhan, line 30: How about renaming the arguments to be more PEP-friendly? Maybe `temperature`, `relative_humidity` or `rel_hum`, `alternate_coeffs`, or so?

src/geocat/comp/heat_index.py
```diff
@@ -0,0 +1,150 @@
...
+    # Input validation on relative humidity
+    if any(rh < 0) or any(rh > 100):
+        raise ValueError('heat_index: invalid values for relative humidity')
+
+    # Check if relative humidity fractional
+    if all(rh < 1):
+        warnings.warn("WARNING: rh must be %, not fractional; All rh are < 1")
+
+    # Default coefficients for (t>=80F) and (40<gh<100)
+    c = [
```
erogluorhan, line 103: Arguments like this may need renaming. `c` says nothing when reading the code.

test/test_heat_index.py
```diff
@@ -0,0 +1,91 @@
...
+
+import dask.array as da
+import dask.distributed as dd
+import numpy as np
+import xarray as xr
+
+# Import from directory structure if coverage test, or from installed
+# packages otherwise
+if "--cov" in str(sys.argv):
+    from src.geocat.comp.heat_index import heat_index
```
erogluorhan, line 12: Please avoid using the script name (i.e. `.heat_index` in this case) when importing in test scripts. This is because we are trying to test against the packages/namespace, not individual scripts. So, use:

`from src.geocat.comp import heat_index`

test/test_heat_index.py
```diff
@@ -0,0 +1,91 @@
...
+else:
+    from geocat.comp.heat_index import heat_index
```
erogluorhan, line 14: Above comment applies here, too.

test/test_heat_index.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class Test_heat_index(unittest.TestCase):
```
erogluorhan, line 17: Please try to write one more test case, if possible, that covers the following `else` case:

image

erogluorhan (changes requested):
Looks great; please address a few inline comments.

src/geocat/comp/wkSpaceTime.py
```diff
@@ -0,0 +1,6 @@
+import numpy
+import xarray
+
+
+def wkSpaceTime():
```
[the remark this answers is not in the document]
anissa111: Oops, it was pulled over from another branch by mistake, deleted!

src/geocat/comp/heat_index.py
```diff
@@ -0,0 +1,150 @@
+import warnings
+
+import numpy as np
+import xarray as xr
+
+from geocat.comp import _is_duck_array
```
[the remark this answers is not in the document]
anissa111: You're right, I'd prefer to not include it in the geocat.comp namespace. Would not including it allow us to remove it from the internal api documentation, as well?

test/test_heat_index.py
anissa111: Right, thank you! I did it that was initially because of some namespace issues, but it appears to have solved itself.

src/geocat/comp/heat_index.py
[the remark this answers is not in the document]
anissa111: Okay, done!

test/test_heat_index.py
anissa111: Done!

michaelavs (approved):
This looks good to me!

src/geocat/comp/heat_index.py
```diff
@@ -0,0 +1,158 @@
...
+    # convert inputs to numpy arrays if necessary
+    if not _is_duck_array(temperature):
+        temperature = np.atleast_1d(temperature)
+    if not _is_duck_array(relative_humidity):
+        relative_humidity = np.atleast_1d(relative_humidity)
+
+    # check to ensure dimensions of inputs not greater than 1
+    if temperature.ndim > 1 or relative_humidity.ndim > 1:
+        raise ValueError('heat_index: inputs must have at most one dimension')
+
```
andersy005, line 96: @anissa111, 

I am curious about why this computation works on 1d arrays only. if I understand correctly, we should be able to compute the heat index as long as `temperature` and `relative_humidity` array shapes are the same/compatible. 

E.g.: Assuming temperature and relative_humidity are 2D arrays: `temperature(lat, lon)` and `relative_humidity(lat, lon)`. This should/would yield `heat_index` with the following dimensionality: `heat_index(lat, lon)`, right? 



If the computation requires 1D arrays, could we accept n-dimensional arrays as inputs and then use [NumPy's ravel()](https://numpy.org/doc/stable/reference/generated/numpy.ravel.html) to create a 1D array view that gets used in the calculation?

anissa111: Thank you for this input! The way the calculation was set up did require 1D arrays. @erogluorhan and I have reworked the calculation to be able to accept multidimensional arrays in the most recent version.

erogluorhan (approved):
It looks great!

PR #144, opened 2021-04-16
Science at Scale Scenario 3 functions

anissa111:
This is a work in progress PR for the implementation of ```psychrometric_constant```,  ```saturation_vapor_pressure```, ```actual_saturation_vapor_pressure```, and ```saturation_vapor_pressure_slope``` (originally ```psychro_fao56```, ```satvpr_temp_fao56```, ```satvpr_tdew_fao56```, and ```satvpr_slope_fao56``` respectively).

These functions are being implemented for the purpose of S@S Scenario 3, described [here](https://github.com/NCAR/science-at-scale/issues/3).


.pre-commit-config.yaml
```diff
@@ -21,7 +16,8 @@ repos:
     - id: trailing-whitespace      # Checks for any tabs or spaces after the last non-whitespace character on the line.
     - id: check-docstring-first    # Checks that code comes after the docstrings.
 
--   repo: https://github.com/PyCQA/isort    # To sort the imports
```
erogluorhan, line 24: Why removing `isort` hook? I have seen the commit about having problems with pre-commit, but please elaborate what the problems were, and if removal was required, etc.

P.S. Whenever a PR makes changes beyond its original goals, please put explanation into the PR description about that change.

erogluorhan (changes requested):
The function and unit test implementations seem to be perfect! Please address a single inline comment plus the unit test behavior below:

All of the unit tests are passing; however, they are giving a lot of "Stream is closed" error, which is likely related to the handling of Dask or Distributed.Client.

.pre-commit-config.yaml
pilotchute: Isort and Yapf were fighting over the imports. This is part of something more that I will bring up at the Friday meeting. for now, we want to have isort removed.

erogluorhan: Got it, thanks! @anissa111 please put this info to first description for future reference.

pilotchute (approved):

.pre-commit-config.yaml
anissa111: will do!

pilotchute:
I may have merged this too soon. The Stream is closed error is due to Dask's thread handling, but isn't an actual problem. 
Worker threads aren't waking up in time to provide frequent enough "alive" pings to the handler. It's unfortunately normal behavior for DASK.

pilotchute:
We can talk about it tomorrow morning.

erogluorhan:
> I may have merged this too soon. The Stream is closed error is due to Dask's thread handling, but isn't an actual problem.
> Worker threads aren't waking up in time to provide frequent enough "alive" pings to the handler. It's unfortunately normal behavior for DASK.

Thanks for explaining "Stream is closed" error. I agree we can talk about it. However, regarding merge, please try not to merge before having all of the reviewers' comments addressed and turned into an approval.

pilotchute:
> > I may have merged this too soon. The Stream is closed error is due to Dask's thread handling, but isn't an actual problem.
> > Worker threads aren't waking up in time to provide frequent enough "alive" pings to the handler. It's unfortunately normal behavior for DASK.
> 
> Thanks for explaining "Stream is closed" error. I agree we can talk about it. However, regarding merge, please try not to merge before having all of the reviewers' comments addressed and turned into an approval.

My bad, was trying to get some things done between meetings, and missed the first half of your change request, just saw "resolved" on the stack below it, and went on down the workflow. Sorry.

PR #145, opened 2021-04-19
Fourier Filters

pilotchute:
Getting close to ready for a pull.


erogluorhan (dismissed):
Will provide a deeper review but I noticed a quick thing for now: Please add docs (internal & user APIs etc.)

pilotchute:
@erogluorhan @anissa111 
Ready for review and merge

erogluorhan:
> Will provide a deeper review but I noticed a quick thing for now: Please add docs (internal & user APIs etc.)

@pilotchute

test/test_fourier_filters.py
```diff
@@ -0,0 +1,282 @@
...
+    t_expected_result = (np.sin(t * m.tau) / 0.1 + np.sin(2 * t * m.tau) / 0.2 +
+                         np.sin(5 * t * m.tau) / 0.5 + np.sin(10 * t * m.tau))
+    t_result = fourier_low_pass(t_data, freq, 15)
+    np.testing.assert_almost_equal(t_result, t_expected_result)
+
+
+def test_one_high_pass():
+    freq = 1000
+    t = np.arange(1000) / freq
+    t_data = (np.sin(t * m.tau) / 0.1 + np.sin(2 * t * m.tau) / 0.2 +
```
erogluorhan, line 33: If I am not missing a nuance, these input data are the same among several test cases. Please define them in a common place to avoid code duplication.

src/geocat/comp/fourier_filters.py
```diff
@@ -0,0 +1,178 @@
...
+def fourier_filter(signal,
+                   frequency,
+                   cutoff_frequency_low=0,
+                   cutoff_frequency_high=0,
+                   time_axis=0,
+                   low_pass=False,
+                   high_pass=False,
+                   band_pass=False,
+                   band_block=False):
+    """Filter a dataset by frequency. This function allowes for low_pass, high_
```
erogluorhan, line 16: What were the NCL functions that are addressed with these Fourier functions? How should we let people know that? Maybe adding some references in the functions' docstrings, or something else?

erogluorhan (changes requested):
The function and unit test implementations are perfect! 

Please address a few inline comments and don't forget adding documentation (user API etc.) though.

src/geocat/comp/fourier_filters.py
```diff
@@ -0,0 +1,178 @@
...
+    -------
+    return_signal : numpy.ndarray, xr.DataArray
+        signal with specified filters applied
+
+    Examples
+    --------
+    Example 1: The tidal cycle needs to be removed from a 10/hr oceanic dataset,
+    (https://tidesandcurrents.noaa.gov/waterlevels.html?id=9415020&units=standard&bdate=20210101&edate=20210131&timezone=GMT&datum=MLLW&interval=6&action=data)
+
+    >>>import xarray as xr
```
anissa111, line 59: There should be a space here for all the docstring examples. Also, you're never importing the fourier function itself in the examples.
```suggestion
    >>> import xarray as xr
```

src/geocat/comp/fourier_filters.py
```diff
@@ -0,0 +1,178 @@
...
+    >>>axs[0].plot(np.real(no_tide)[start:end])
+    >>>axs[1].set_title('imag')
+    >>>axs[1].plot(np.imag(xr_data)[start:end])
+    >>>axs[1].plot(np.imag(no_tide)[start:end])
+    >>>fig.show()
+    """
+    resolution = frequency / len(signal)
+    signal = np.swapaxes(signal, time_axis, 0)
+    res_fft = np.fft.fft(signal, axis=0)
+    cfl_index = m.floor(cutoff_frequency_low / resolution)
```
anissa111, line 109: Is there a reason you're using the math functions instead of the numpy functions here?

src/geocat/comp/fourier_filters.py
```diff
@@ -0,0 +1,178 @@
...
+                   high_pass=False,
+                   band_pass=False,
+                   band_block=False):
+    """Filter a dataset by frequency. This function allowes for low_pass, high_
+    pass, band_pass, or band_block filtering of the data's freqency
+    representation.
+
+    Parameters
+    ----------
+    signal : numpy.ndarray, xr.DataArray,
```
anissa111, line 22: Are we requiring numpy or xarray and not allowing for list input?

anissa111 (changes requested):
This looks good, just a few inline comments below.

Also is there a reason that dask tests were not included in the unittests?

src/geocat/comp/fourier_filters.py
pilotchute: done

src/geocat/comp/fourier_filters.py
pilotchute: No particular reason. math functions are "more native" and sliiiiightly faster. And I used math because I have unresolved java habits. :)

src/geocat/comp/fourier_filters.py
pilotchute: In order to run the fft, we need the data in a numpy array, we could turn a nested list into a numpy array, and then call to_list later. But I really do not expect anyone to ever call this function against a list.

docs/internal_api/index.rst
```diff
@@ -27,6 +27,16 @@ GeoCAT-comp routines
 
    geocat.comp.eofunc._generate_eofs_solver
 
+   geocat.comp.fourier_filters.fourier_bank_block
```
erogluorhan, line 30: Let us not list these functions in the internal API and only list them in user API to avoid any confusion. This is because they all are in the namespace and can be called by the user, and there is no specific internal usage of them (even though `fourier_filter` can be used as an internal function by the others).

docs/user_api/index.rst
```diff
@@ -28,6 +28,16 @@ GeoCAT-comp routines
 
    geocat.comp.eofunc_pcs
 
+   geocat.comp.fourier_filters.fourier_bank_block
```
erogluorhan, line 31: Typo: `_bank_block`

pilotchute: thanks, I was having trouble finding the error.

src/geocat/comp/fourier_filters.py
erogluorhan: Per our discussion in the meeting, I think we are good with this as long as we add  usage examples for `smthClmDay` and `rmMonAnnCyc` functions under GeoCAT-examples. I do not have a strong preference about whether to put a statement into the docstrings here or not to put.

erogluorhan (approved):
Looks great to me!

src/geocat/comp/fourier_filters.py
pilotchute: I don't think a statement belongs in the docstring, but my opinion isn't strong.

erogluorhan: I agree

pilotchute:
@anissa111 Would you re-review, are we ready to merge this?

anissa111 (approved):
Looks good!

PR #151, opened 2021-05-04
bugfix

pilotchute:

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -12,4 +12,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - uses: pre-commit/action@v2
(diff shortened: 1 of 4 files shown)


erogluorhan (approved):

erogluorhan:
@pilotchute I approved but have a question: You are obligating `numpy>=1.20.2` for GeoCAT-f2py but are not obligating Numpy to be greater than any version here. Do you think it would for without any issue or should we obligate `numpy>=1.20.2` here, too?

pilotchute:
> @pilotchute I approved but have a question: You are obligating `numpy>=1.20.2` for GeoCAT-f2py but are not obligating Numpy to be greater than any version here. Do you think it would for without any issue or should we obligate `numpy>=1.20.2` here, too?

This should be the cleanest solution. since it is pinned in the geocat-f2py dependancy, there's no risk of pulling the wrong version, or having a conflict.

PR #155, opened 2021-05-26
Apply a Workaround for Metpy's Issue with 4D Xarray in interp_hybrid_to_sigma 

erogluorhan:
This PR applies a workaround to `interp_hybrid_to_pressure` function for the issue with `metpy`'s xarray interface:
   
DESCRIPTION: `metpy.interpolate.interpolate_1d` had "no implementation found for 'numpy.apply_along_axis'" error for cases where the input is `xarray.Dataarray` and has more than 3 dimensions (e.g. 4th dim of `time`).

The workaround used in this PR uses `dask.array.core.map_blocks` instead of `xarray.apply_ufunc` and auto-chunks input arrays (if unchunked) to ensure using only `Numpy` interface of `metpy.interpolate.interpolate_1d` (i.e. When call map_blocks with Dask array, it calls Numoy interface of metpy function).

### DIFF
--- a/src/geocat/comp/interp_hybrid_to_pressure.py
+++ b/src/geocat/comp/interp_hybrid_to_pressure.py
@@ -120,20 +120,15 @@ def interp_hybrid_to_pressure(data,
     #     }},
     # )
 
-    if isinstance(data, xr.DataArray):
-        # If an unchunked Xarray input is given, chunk it just with its dims
-        if (data.chunks is None):
-            data_chunk = dict([
-                (k, v) for (k, v) in zip(list(data.dims), list(data.shape))
-            ])
-            data = data.chunk(data_chunk)
+    # If an unchunked Xarray input is given, chunk it just with its dims
+    if (data.chunks is None):
+        data_chunk = dict([
+            (k, v) for (k, v) in zip(list(data.dims), list(data.shape))
+        ])
+        data = data.chunk(data_chunk)
 
-        if (pressure.chunks is None):
-            pressure_chunk = dict([
-                (k, v)
-                for (k, v) in zip(list(pressure.dims), list(pressure.shape))
-            ])
-            pressure = pressure.chunk(pressure_chunk)
+    # Chunk pressure equal to data's chunks
+    pressure = pressure.chunk(data.chunks)
 
     # Output data structure elements
     out_chunks = list(data.chunks)
--- a/src/geocat/comp/interp_hybrid_to_pressure.py
+++ b/src/geocat/comp/interp_hybrid_to_pressure.py
@@ -121,7 +121,7 @@ def interp_hybrid_to_pressure(data,
     # )
 
     # If an unchunked Xarray input is given, chunk it just with its dims
-    if (data.chunks is None):
+    if data.chunks is None:
         data_chunk = dict([
             (k, v) for (k, v) in zip(list(data.dims), list(data.shape))
         ])
--- a/src/geocat/comp/interp_hybrid_to_pressure.py
+++ b/src/geocat/comp/interp_hybrid_to_pressure.py
@@ -162,7 +162,7 @@ def interp_hybrid_to_pressure(data,
         data.dims[i] if i != interp_axis else "plev" for i in range(data.ndim)
     ]
 
-    # This is needed with the existence of above workaround block
+    # Rename output dims. This is only needed with above workaround block
     dims_dict = {output.dims[i]: dims[i] for i in range(len(output.dims))}
     output = output.rename(dims_dict)
 
--- a/src/geocat/comp/interp_hybrid_to_pressure.py
+++ b/src/geocat/comp/interp_hybrid_to_pressure.py
@@ -96,8 +96,8 @@ def interp_hybrid_to_pressure(data,
     # For the issue with metpy's xarray interface:
     #
     # `metpy.interpolate.interpolate_1d` had "no implementation found for
-    # 'numpy.apply_along_axis'" for cases where the input is xarray.Dataarray
-    # and has more than 3 dimensions (e.g. 4th dim of `time`).
+    # 'numpy.apply_along_axis'" issue for cases where the input is
+    # xarray.Dataarray and has more than 3 dimensions (e.g. 4th dim of `time`).
 
     # Use dask.array.core.map_blocks instead of xarray.apply_ufunc and
     # auto-chunk input arrays to ensure using only Numpy interface of
@@ -107,13 +107,13 @@ def interp_hybrid_to_pressure(data,
     # # Apply Dask parallelization with xarray.apply_ufunc
     # output = xr.apply_ufunc(
     #     _vertical_remap,
-    #     data.values,
-    #     pressure.values,
+    #     data,
+    #     pressure,
     #     exclude_dims=set((lev_dim,)),  # Set dimensions allowed to change size
     #     input_core_dims=[[lev_dim], [lev_dim]],  # Set core dimensions
     #     output_core_dims=[["plev"]],  # Specify output dimensions
     #     vectorize=True,  # loop over non-core dims
-    #     # dask="parallelized",  # Dask parallelization
+    #     dask="parallelized",  # Dask parallelization
     #     output_dtypes=[data.dtype],
     #     dask_gufunc_kwargs={"output_sizes": {
     #         "plev": len(new_levels)
--- a/src/geocat/comp/interp_hybrid_to_pressure.py
+++ b/src/geocat/comp/interp_hybrid_to_pressure.py
@@ -1,7 +1,6 @@
 import cf_xarray
 import metpy.interpolate
 import numpy as np
-import xarray
 import xarray as xr
 
 __pres_lev_mandatory__ = np.array([
@@ -150,7 +149,7 @@ def interp_hybrid_to_pressure(data,
     # End of Workaround
     ###############################################################################
 
-    output = xarray.DataArray(output)
+    output = xr.DataArray(output)
 
     # Set output dims and coords
     dims = [
--- a/src/geocat/comp/interp_hybrid_to_pressure.py
+++ b/src/geocat/comp/interp_hybrid_to_pressure.py
@@ -1,6 +1,7 @@
 import cf_xarray
 import metpy.interpolate
 import numpy as np
+import xarray
 import xarray as xr
 
 __pres_lev_mandatory__ = np.array([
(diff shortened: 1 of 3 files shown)


src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -86,26 +90,76 @@ def _vertical_remap(data, pressure):
 
         return func_interpolate(new_levels, pressure, data, axis=interp_axis)
```
dcherian, line 91:
```suggestion
        return func_interpolate(new_levels, pressure, data, axis=-1)
```

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -86,26 +90,76 @@ def _vertical_remap(data, pressure):
...
 
-    # Apply vertical interpolation
-    # Apply Dask parallelization with xarray.apply_ufunc
-    output = xr.apply_ufunc(
+    ###############################################################################
+    # Workaround
+    #
+    # For the issue with metpy's xarray interface:
```
dcherian, line 96: I'm confused by this PR.

If you use `apply_ufunc`, the function `_vertical_remap` will receive numpy arrays, not xarrays, so this code is not using the xarray interface AFAICT. In theory, my suggestion above should fix the old code: `apply_ufunc` moves the core dimension to the end, so you should always pass `axis=-1` to any function that expects it.

Second, the metpy function can handle nD arrays since it is using `apply_along_axis`, so you can avoid the `vectorize=True` bit and get a speedup.

The "no implementation error found" is interesting... I don't see how that is triggered. Can you post a full traceback?

erogluorhan: Hi @dcherian Thanks for your comments! I will look into them when I have some time. In the meantime, please see [this MetPy issue #1889](https://github.com/Unidata/MetPy/issues/1889) I created for Metpy for a minimal example code for this case, if you haven't done so.

erogluorhan: Hi @dcherian ,

I had a chance to work a bit further on this issue. Here are my findings:

- You are right, `apply_ufunc` is delivering `ndarray`s to `_vertical_remap`, thanks! 
- However, there is still a difference in `metpy` decorators where they call `log_interpolate_1d`. Regardless of you call either via `apply_ufunc` or directly `metpy.interpolate.log_interpolate_1d` (as I did in [this MetPy issue #1889 ](https://github.com/Unidata/MetPy/issues/1889) ), the input data type you are giving (i.e. numpy or xarray) is ending up in different data types handled within metpy decorators. I observed that in `log_xp` in `interpolate_1d(log_x, log_xp, *args, axis=axis, fill_value=fill_value)` was `Array` when I gave `ndarray` inputs but `Quantity` when I gave `xarray.DataArray` inputs, and I got the following traceback when it was `Quantity`:

```
Traceback (most recent call last):
  File "/Users/oero/miniconda3/envs/geocat_comp_build/lib/python3.9/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "/Users/oero/src/geocat-comp/test/test_interp_hybrid_to_pressure.py", line 104, in <module>
    a.test_interp_hybrid_to_pressure_atmos()
  File "/Users/oero/src/geocat-comp/test/test_interp_hybrid_to_pressure.py", line 48, in test_interp_hybrid_to_pressure_atmos
    u_int = interp_hybrid_to_pressure(_data,
  File "/Users/oero/miniconda3/envs/geocat_comp_build/lib/python3.9/site-packages/geocat/comp/interp_hybrid_to_pressure.py", line 110, in interp_hybrid_to_pressure
    output_xr = metpy.interpolate.log_interpolate_1d(new_levels, pressure, data, axis=interp_axis)
  File "/Users/oero/miniconda3/envs/geocat_comp_build/lib/python3.9/site-packages/metpy/xarray.py", line 1206, in wrapper
    result = func(*bound_args.args, **bound_args.kwargs)
  File "/Users/oero/miniconda3/envs/geocat_comp_build/lib/python3.9/site-packages/metpy/interpolate/one_dimension.py", line 226, in log_interpolate_1d
    return interpolate_1d(log_x, log_xp, *args, axis=axis, fill_value=fill_value)
  File "/Users/oero/miniconda3/envs/geocat_comp_build/lib/python3.9/site-packages/metpy/xarray.py", line 1206, in wrapper
    result = func(*bound_args.args, **bound_args.kwargs)
  File "/Users/oero/miniconda3/envs/geocat_comp_build/lib/python3.9/site-packages/metpy/interpolate/one_dimension.py", line 127, in interpolate_1d
    minv = np.apply_along_axis(np.searchsorted, axis, xp, x[sort_x])
  File "<__array_function__ internals>", line 5, in apply_along_axis
TypeError: no implementation found for 'numpy.apply_along_axis' on types that implement __array_function__: [<class 'pint.quantity.build_quantity_class.<locals>.Quantity'>]
``` 

- `axis=-1` did not work either. It is not surprising to me, because this does not seem to be an error with `apply_ufunc` and its handling of the core dimensions. Even calling the metpy function with xarray inputs without `apply_ufunc` (as I did in the above issue) ends up with the same failure.

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -1,6 +1,7 @@
 import cf_xarray
 import metpy.interpolate
 import numpy as np
+import xarray
 import xarray as xr
```
anissa111, line 5:
```suggestion
```
We don't need to import xarray aliased and directly

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -1,6 +1,7 @@
 import cf_xarray
```
anissa111, line 1:
```suggestion
```
unused import

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -1,6 +1,7 @@
...
 import metpy.interpolate
```
anissa111, line 2: metpy needs to be added to the project requirements

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -86,26 +90,76 @@ def _vertical_remap(data, pressure):
...
+    #     vectorize=True,  # loop over non-core dims
+    #     dask="parallelized",  # Dask parallelization
+    #     output_dtypes=[data.dtype],
+    #     dask_gufunc_kwargs={"output_sizes": {
+    #         "plev": len(new_levels)
+    #     }},
+    # )
+
+    # If an unchunked Xarray input is given, chunk it just with its dims
+    if (data.chunks is None):
```
anissa111, line 124:
```suggestion
    if data.chunks is None:
```
redundant parenthesis

src/geocat/comp/interp_hybrid_to_pressure.py
```diff
@@ -86,26 +90,76 @@ def _vertical_remap(data, pressure):
...
+    #     exclude_dims=set((lev_dim,)),  # Set dimensions allowed to change size
+    #     input_core_dims=[[lev_dim], [lev_dim]],  # Set core dimensions
+    #     output_core_dims=[["plev"]],  # Specify output dimensions
+    #     vectorize=True,  # loop over non-core dims
+    #     dask="parallelized",  # Dask parallelization
+    #     output_dtypes=[data.dtype],
+    #     dask_gufunc_kwargs={"output_sizes": {
+    #         "plev": len(new_levels)
+    #     }},
+    # )
```
anissa111, line 121:
```suggestion
```
Do we need to keep the old version in the comments?

anissa111 (commented):
This makes sense to me, thank you for the work around. I left some minor comments below.

src/geocat/comp/interp_hybrid_to_pressure.py
andersy005: This import is actually required for the `.cf` accessor to work:

https://github.com/NCAR/geocat-comp/blob/eae0d27f353db2bfae9ae36ad2b615fdca66cf44/src/geocat/comp/interp_hybrid_to_pressure.py#L64

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: We don't have to but could better keep it at least for a few versions in the near future since this is the only function we applied `xarray.apply_ufunc` AFAIK. The issue here does not seem to be due to `apply_ufunc`; instead, it seems to be of `metpy.interpolate_1d` issue. I couldn't figure yet why metpy's interpolate working nicely with `map_blocks` but it worked. I opened [an issue](https://github.com/Unidata/MetPy/issues/1889) at metpy repo; let us see if it gets any response. I am recommending to keep the commented block until then.

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Good catch, thanks! I removed `import xarray` and converted references to it in the code to alias instead.

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: I agree with @andersy005 , it is used, though it is confusing to see the `import cf_xarray` as if it is unused in PyCharm interface.

src/geocat/comp/interp_hybrid_to_pressure.py
erogluorhan: Great catch! Added `metpy` to requirements.txt

src/geocat/comp/interp_hybrid_to_pressure.py
anissa111: Okay, that makes sense to me!

pilotchute:
merging once checks complete

src/geocat/comp/interp_hybrid_to_pressure.py
pilotchute: @erogluorhan what are your thoughts on this proposed change?

src/geocat/comp/interp_hybrid_to_pressure.py
pilotchute: @dcherian does this conversation resolve your comment?

dcherian: > I observed that in log_xp in interpolate_1d(log_x, log_xp, *args, axis=axis, fill_value=fill_value) was Array when I gave ndarray inputs

So not `ndarray` even though you provide `ndarray` inputs. When you eventually compute, that function will receive ndarrays even with your map_blocks approach, so it should work (since your tests pass!). :) I'm still confused... but maybe this should just get merged.

PR #156, opened 2021-06-04
Month_to_season12 and month_to_seasonN

hCraker:
This PR will create the remaining functions in the month_to_season family of [climatology functions](https://www.ncl.ucar.edu/Document/Functions/climo.shtml). 

- [X] month_to_season (already implemented see #154)
- [x] month_to_season12
- [ ] month_to_seasonN

This is one PR that addresses issue #66

Issue #66, opened 2020-04-20
Add support for climatology (means)


clyne:
Provide support for general climatologies of a user-specified frequency (e.g. monthly, seasonal). Specifically, provide equivalent functionality to NCL's clm* functions. See https://www.ncl.ucar.edu/Document/Functions/climo.shtml

This capability will be provided by refactoring the ESMLab climatology function: https://esmlab.readthedocs.io/en/latest/api.html#esmlab.core.climatology

API and capability changes to ESMLab  core.climatology:

+ Operator should work on Xarray DataArray objects, not Xarray DataSets

+ Time handling functions should be extracted for standalone use (a prototype time handling utility exists here: https://github.com/NCAR/xtimeutil)

+ Add support for seasonal frequency (See NCL's month_to_season functions)

Notes:

+ Xarray should not be relied on for time decoding

+ Current code relies on "time bounds", which may not be present in a data set

+ Possible useful example jupyter notebook:

https://nbviewer.jupyter.org/github/andersy005/brouillons-quotidien/blob/master/climatology/clim.ipynb

Keith Lindsey is a resource for science questions

pilotchute:
https://www.ncl.ucar.edu/Document/Functions/climo.shtml

Contains a list of climatology functions to port.

erogluorhan:
@pilotchute @andersy005 Please schedule a time to sit together and discuss how to migrate Esmlab functions into GeoCAT-comp. Such a migration could be performed in one of the coming monthly releases I believe.

anissa111:
@pilotchute @hCraker, should this issue be closed?

hCraker:
@anissa111 
The original request was for multiple climatology functions which includes everything on [this list](https://www.ncl.ucar.edu/Document/Functions/climo.shtml). There was a particular focus on recreating the clm* functions and month to season. The clm* functions were completed and deployed in PR #158.

The month_to_season functions were partially implemented in that same PR. Averages are calculated for the 4 meteorological seasons DJF, MAM, JJA, and SON, but the user cannot specify different 3 month seasons yet like in NCL.

This original issue encompassed a lot of things. I think it would be best to close it and open new issues for the remaining climatology functions.

kafitzgerald:
At this point much of this (if not all) seems to have been implemented and the original code and notebook links are no longer valid.

Closing this out in favor of tracking NCL to Python progress on climatology functions via geocat-applications.

PR #158, opened 2021-06-08
Averaging time series data

hCraker:
This PR adds a general function to compute sequential and rolling averages for `xarray.DataArray`, `xarray.Dataset`, and `numpy.ndarray` data that vary in time. This PR also provides novel functions and wrappers of the general one to calculate climatological averages for various time periods which are listed below:

- [x] hourly
- [x] daily
- [x] ~~weekly~~ (not needed and can't reliably be done given how weeks change from year to year)
- [x] monthly
- [x] seasonally
- [x] yearly

Note: The implementation of this general function may change PR #156. Consider that PR on hold until this PR is complete.

This function addresses some of the functionality requested in #66

### EDIT:
This function and the problem it was trying to solve, have changed drastically during the course of development. For the record, `climatology_average` calculates climatological averages for the time periods listed above and `calendar_average` calculates the average for each period (i.e. the monthly average for every January from 2011-2015 versus the January climatological average across those 5 years)


src/geocat/comp/climatology.py
```diff
@@ -322,3 +326,144 @@ def month_to_season(
...
+    computed_dset: same type as dset
+        The computed monthly means
+    """
+    # TODO: check if data has correct time resolution
+    time_dim = _get_time_coordinate_info(dset, time_dim)
+    if across_years:
+        return dset.groupby(time_dim + '.month').mean()
+    else:
+        return dset.groupby(time_dim + '.year').map(_avg_groups,
+                                                    group=time_dim + '.month')
```
andersy005, line 469: @hCraker, Could we avoid code duplication here by merging these two functions under one function. I am imagining something along these lines:

```python
def average(dset, freq, time_dim=None, across_years=True):
    time_dim = _get_time_coordinate_info(dset, time_dim)
    freqs = {'daily': 'dayofyear', 'monthly': 'month'} # This could be extended to accommodate other frequencies if need be. 
    group = f'{time_dim}.{freqs[freq]}'
    if across_years:
        return dset.groupby(group).mean() 
    else:
        return dset.groupby(f'{time_dim}.year').map(_avg_groups, group=group)
```

src/geocat/comp/climatology.py
```diff
@@ -322,3 +326,144 @@ def month_to_season(
...
+    -------
+    computed_dset: same type as dset
+        The computed data
+    """
+
+    min_periods = rolling_kwargs.get('min_periods')
+    center = rolling_kwargs.get('center')
+
+    if isinstance(dset, np.ndarray):  # TODO: duck typing for arrays
+        if time_dim == None or not isinstance(time_dim, int):
```
andersy005, line 362: Use `time_dim` is None rather than `time_dim == None`
```suggestion
        if time_dim is None or not isinstance(time_dim, int):
```

src/geocat/comp/climatology.py
andersy005: 👍🏽 Sounds good!

src/geocat/comp/climatology.py
pilotchute: @andersy005 this is a draft PR from my intern, she and I will do a code review before we take this out of draft, and then we will request reviews once it is a non-draft PR. 

Until then, she will be actively developing, and there's no need to comment on the occasional weirdness that happens in early draft phases.

dcherian:
I see this is a draft but you should take a look at [coarsen](https://xarray.pydata.org/en/stable/user-guide/computation.html#coarsen-large-arrays). 

It also looks a lot like you're reimplementing [resample](https://xarray.pydata.org/en/stable/user-guide/time-series.html#resampling-and-grouped-operations) but I am not certain. If I am misunderstanding things, I apologize for the noise.

hCraker:
> I see this is a draft but you should take a look at [coarsen](https://xarray.pydata.org/en/stable/user-guide/computation.html#coarsen-large-arrays).
> 
> It also looks a lot like you're reimplementing [resample](https://xarray.pydata.org/en/stable/user-guide/time-series.html#resampling-and-grouped-operations) but I am not certain. If I am misunderstanding things, I apologize for the noise.

Hi @dcherian. I'm not well versed in xarray, so I appreciate you pointing out more obscure functions. Here's what I'm trying to do. For the `daily_avg` function, I want to take in sub daily data, group it by year, then by month, and then by day. I want to average the data in those groups and then aggregate the averages back into one `xr.DataArray` or `xr.Dataset` (depending on the type of the the input). This should return the mean of the data from each day. Users will also have the option of grouping data by month and day (but not by year) to produce climatological averages for each day over a span of years. 

I tried `resample` at one point, and it wasn't working the way I wanted. I can't remember why I abandoned that approach, but I will look at it and `coarsen` again to be sure I'm not reimplementing functionality.

dcherian:
>  I want to take in sub daily data, group it by year, then by month, and then by day.

Is your output is a time series with time vector : `"2001-01-01", "2001-01-02", "2001-02-03"` etc? If so that *should* be `.resample(time="D").mean()`

> Users will also have the option of grouping data by month and day (but not by year) to produce climatological averages for each day over a span of years.

OK this is interesting. But would `.groupby("time.dayofyear")` solve your issue? (see https://xarray.pydata.org/en/stable/user-guide/time-series.html#datetime-components for more). In general, `groupby` is really flexible because you can get quite creative with creating appropriate group labels. One could imagine using `time.dt.strftime` to create a time series of strings formatted as `"month-day"` and grouping by that.

hCraker:
> Is your output is a time series with time vector : `"2001-01-01", "2001-01-02", "2001-02-03"` etc? If so that _should_ be `.resample(time="D").mean()`

This worked! I was using `time="time.day"` which gave me the day of the month and not the full date like I wanted. Thanks for the help.

> But would `.groupby("time.dayofyear")` solve your issue?

I have tried that, but unfortunately it doesn't differentiate between the 60th day of the year sometimes being Feb 29th and sometimes being Mar 1st. I'm going to talk with some potential users to ask them if it would be best to filter out Feb 29th data altogether and not consider leap days, or if they have another way of dealing with this leap year problem.

dcherian:
> One could imagine using time.dt.strftime to create a time series of strings formatted as "month-day" and grouping by that.

This approach would preserve `Feb-29` and `Mar-01` being different.

hCraker:
> > One could imagine using time.dt.strftime to create a time series of strings formatted as "month-day" and grouping by that.
> 
> This approach would preserve `Feb-29` and `Mar-01` being different.

I misunderstood that functionality when I first read your suggestion since I'm not familiar with `DatetimeAccessors` and `strftime`, but it groups the data the way I want it to.

Thank you so, so much! There is a lot more to xarray than I originally thought, and examples like this one that show more nuanced ways of grouping data should definitely be on the xarray time series documentation page. The rest of this PR will go by quickly now that I know I can group things this way. 🎉

dcherian:
> examples like this one that show more nuanced ways of grouping data should definitely be on the xarray time series documentation page. 

PRs are always very welcome! =)

src/geocat/comp/climatology.py
hCraker: FYI I added this suggestion as part of one of my other commits. This can be labeled as resolved now.

hCraker:
In my opening comment for this PR, I said that `np.ndarray` would be supported. I have decided against doing that for these functions since there seems to be little need. I have yet to find a user that doesn't already convert their data into an `xr.Dataset` or `xr.DataArray`. Additionally, computing weekly climatology averages is impossible to do without making a ton of assumptions about how weeks are defined, and there also seems to be no need for this functionality. A 7 day rolling average is the closest functionality that it likely to be desired by users.

Once I'm done writing the remaining test cases and clean up documentation, this will be ready to review.

src/geocat/comp/climatology.py
```diff
@@ -322,3 +324,142 @@ def month_to_season(
...
+            dset = dset.resample({
+                time_dim: frequency
+            }, loffset=offset).mean().dropna(time_dim)
+
+            # Compute the weights for the months in each season so that the
+            # seasonal averages account for months being of different lengths
+            month_length = dset[time_dim].dt.days_in_month.groupby(time_dim +
+                                                                   '.season')
+            weights = month_length / month_length.sum()
+            dset = (dset * weights).groupby(time_dim + '.season')
```
andersy005, line 425:
```suggestion
            dset = (dset * weights).groupby(f"{time_dim}.season")
```

src/geocat/comp/climatology.py
```diff
@@ -322,3 +324,142 @@ def month_to_season(
...
+                    {'strftime': time_dim})
+
+            # Create array of datetimes to set as time coordinate of returned data
+            # Offsets are used to ensure the time coordinate of the returned climatology is centered on the period
+            start_day = dset[time_dim].values[0]
+            end_day = dset[time_dim].values[-1]
+
+            time = pd.date_range(f'{median_yr:.0f}-{start_day}',
+                                 f'{median_yr:.0f}-{end_day}',
+                                 freq=frequency) + offset_obj
```
andersy005, line 443: Since pandas' calendar support is limited to `gregorian` calendar, using `pd.date_range` here makes this functionality unusable for model output datasets with non-standard calendars (`noleap`, `360_day`, `all_leap`, `365_day`, etc). Could we use [`xr.cftime_range`](https://xarray.pydata.org/en/stable/generated/xarray.cftime_range.html) instead? Maybe this belongs in a follow-up PR...

src/geocat/comp/climatology.py
```diff
@@ -322,3 +324,142 @@ def month_to_season(
...
+def calendar_average(
+        dset: typing.Union[xr.DataArray, xr.Dataset],
+        freq: str,
+        time_dim: str = None,
+        climatology: bool = False) -> typing.Union[xr.DataArray, xr.Dataset]:
+    """This function computes averages according to a given time frequency.
+
+    Parameters
+    ----------
+    dset : `xarray.Dataset`, `xarray.DataArray`
```
andersy005, line 338: I don't think the backticks are needed when using numpydoc style 

```suggestion
    dset : xarray.Dataset, xarray.DataArray
```

src/geocat/comp/climatology.py
```diff
@@ -322,3 +324,142 @@ def month_to_season(
...
+
+    freq : str
+        Frequency alias. Accepted alias:
+            - 'hour': for hourly averages
+            - 'day': for daily averages
+            - 'month': for monthly averages
+            - 'season': for meteorological seasonal averages (DJF, MAM, JJA, and SON)
+            - 'yearly': for yearly averages
+
+    time_dim : `str`, Optional
```
andersy005, line 349:
```suggestion
    time_dim : str, Optional
```

src/geocat/comp/climatology.py
```diff
@@ -322,3 +324,142 @@ def month_to_season(
...
+        Name of the time coordinate for `xarray` objects
+
+    climatology : `boolean`
```
andersy005, line 352:
```suggestion
    climatology : bool
```

src/geocat/comp/climatology.py
```diff
@@ -322,3 +324,142 @@ def month_to_season(
...
+        raise KeyError(
+            f"contributed: calendar_average: bad period: PERIOD = {freq}. Valid periods include: {list(freq_dict.keys())}"
+        )
+
+    # If freq is 'season', key is set to monthly in order to calculate monthly
+    # averages which are then used to calculate seasonal averages
+    if freq == 'season' or freq == 'year':
+        key = 'month'
+    else:
+        key = freq
```
andersy005, line 392:
```suggestion
    key = 'month' if freq in {'season', 'year'} else freq
```

andersy005 (commented):
Thank you for working on this, @hCraker! I left a few inline comments

Ccing @dcherian who may have additional feedback

src/geocat/comp/climatology.py
```diff
@@ -322,3 +324,142 @@ def month_to_season(
...
+        if freq == 'season':
+            # Calculate monthly average before calculating seasonal climatologies
+            dset = dset.resample({
+                time_dim: frequency
+            }, loffset=offset).mean().dropna(time_dim)
+
+            # Compute the weights for the months in each season so that the
+            # seasonal averages account for months being of different lengths
+            month_length = dset[time_dim].dt.days_in_month.groupby(time_dim +
+                                                                   '.season')
```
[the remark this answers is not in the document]
hCraker: I'm going to commit this suggestion, but could you explain why f-string would be better than concatenation? Is it faster?

src/geocat/comp/climatology.py
hCraker: Backticks aren't required, but if you specify the argument type in the docstring like this ```:class: `xarray.DataArray` ``` it is turned into a hyperlink that leads to the documentation page for that class, object, etc. I just forgot to add `:class:` before it and `:obj:` before `str` and `boolean`. Currently the documentation pages for geocat-comp are inconsistent in whether or not they use this feature. I'm hoping to go through and rectify that at some point.

src/geocat/comp/climatology.py
hCraker: This is much cleaner. Thanks!

src/geocat/comp/climatology.py
hCraker: At the moment, I can't use `xr.cftime_range` because it is incompatible with `pd.offsets`. This makes it so I cannot shift the time coordinate to the center of the averaged period. I'm currently looking for an alternative, but I don't think this functionality exists in `xarray`. I'll keep looking, but I may end up removing the date shifting so this code works with different calendars since that seems more important.

andersy005: I see... How about bypassing `pd.offset` by using `cftime_range()` to generate **time bounds** and then computing the average of these bounds to get the data ticks. It's my understanding that this would shift the time data tick to be in the `middle` of the bounds interval (which is what I think you are doing with the date shifting, right?). 

```python
In [36]: time = pd.date_range('2020-01', '2020-12', freq='MS')

In [37]: time
Out[37]: 
DatetimeIndex(['2020-01-01', '2020-02-01', '2020-03-01', '2020-04-01',
               '2020-05-01', '2020-06-01', '2020-07-01', '2020-08-01',
               '2020-09-01', '2020-10-01', '2020-11-01', '2020-12-01'],
              dtype='datetime64[ns]', freq='MS')

In [38]: time + offset_obj
Out[38]: 
DatetimeIndex(['2020-01-15', '2020-02-15', '2020-03-15', '2020-04-15',
               '2020-05-15', '2020-06-15', '2020-07-15', '2020-08-15',
               '2020-09-15', '2020-10-15', '2020-11-15', '2020-12-15'],
              dtype='datetime64[ns]', freq=None)
```

The equivalent of this with `cftime` would look like this 

```python
In [39]: cftimes_bounds = xr.cftime_range('2020-01', '2021-01', freq='MS')
In [52]: cftimes_bounds.shape
Out[52]: (13,)
```


```python
In [51]: xr.DataArray(np.vstack((cftimes[:-1], cftimes[1:])).T, dims=['time', 'nbd']).mean(dim='nbd')
Out[51]: 
<xarray.DataArray (time: 12)>
array([cftime.DatetimeGregorian(2020, 1, 16, 12, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 2, 15, 12, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 3, 16, 12, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 4, 16, 0, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 5, 16, 12, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 6, 16, 0, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 7, 16, 12, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 8, 16, 12, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 9, 16, 0, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 10, 16, 12, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 11, 16, 0, 0, 0, 0),
       cftime.DatetimeGregorian(2020, 12, 16, 12, 0, 0, 0)], dtype=object)
Dimensions without coordinates: time
```

andersy005: Handling different calendars becomes a matter of getting the calendar information from user's input:

```python
In [53]: cftimes_bounds = xr.cftime_range('2020-01', '2021-01', freq='MS', calendar='360_day')

In [54]: xr.DataArray(np.vstack(( cftimes_bounds[:-1],  cftimes_bounds[1:])).T, dims=['time', 'nbd']).mean(dim='nbd')
Out[54]: 
<xarray.DataArray (time: 12)>
array([cftime.Datetime360Day(2020, 1, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 2, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 3, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 4, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 5, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 6, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 7, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 8, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 9, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 10, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 11, 16, 0, 0, 0, 0),
       cftime.Datetime360Day(2020, 12, 16, 0, 0, 0, 0)], dtype=object)
Dimensions without coordinates: time
```

hCraker: >It's my understanding that this puts the time data tick in the `middle` of the bounds interval (which is what I think you are doing with the date shifting, right?).

You're right that I am trying to shift the time data to the middle of the interval, but I want to do so rounded to the nearest `day`, `hour`, etc based on the resolution of the original time dimension. The `cftime` implementation in your first example results in date times with an hour attribute even though the original data has a resolution of days. Despite that, I think this will work for now. I can try using `floor` on the data to get rid of the hours.

I'm not quite sure how end users will want the time coordinate transformed after computing the averages, but I'll show them this and ask when I talk to users next week. Thanks for the suggestion!

hCraker: In the example you posted, the data goes from January 2020 to December 2020. But your call to `cftime_range` goes from January 2020 to January 2021. I'm trying to find a way extract the last datetime from the dataset and add one month to it so I can pass that into `cftime_range`. Unfortunately I'm finding this difficult since timedeltas don't support adding a month to a datetime. @andersy005 do you have any ideas of how I can generalize what you suggested?

andersy005: @hCraker, you can use the `start` and `end` dates from the dataset to generate the ranges, and then use the `shift` method in conjunction with the append method to generate the `final` ranges. It's a bit hacky but it works :) 


- Generate the ranges using `start` and `end` dates from the dataset

```python
In [36]: start, end, freq = '2020-01', '2020-12', 'MS'

In [37]: x = xr.cftime_range(start, end, freq=freq)

In [38]: x.shape
Out[38]: (12,)

In [39]: x
Out[39]: 
CFTimeIndex([2020-01-01 00:00:00, 2020-02-01 00:00:00, 2020-03-01 00:00:00,
             2020-04-01 00:00:00, 2020-05-01 00:00:00, 2020-06-01 00:00:00,
             2020-07-01 00:00:00, 2020-08-01 00:00:00, 2020-09-01 00:00:00,
             2020-10-01 00:00:00, 2020-11-01 00:00:00, 2020-12-01 00:00:00],
            dtype='object', length=12, calendar='gregorian', freq='MS')

```

- Use `.shift()` + `.append()`

```python
In [40]: y = x.append(x.shift(1, freq=freq)[-1:])

In [41]: y.shape
Out[41]: (13,)

In [42]: y
Out[42]: 
Index([2020-01-01 00:00:00, 2020-02-01 00:00:00, 2020-03-01 00:00:00,
       2020-04-01 00:00:00, 2020-05-01 00:00:00, 2020-06-01 00:00:00,
       2020-07-01 00:00:00, 2020-08-01 00:00:00, 2020-09-01 00:00:00,
       2020-10-01 00:00:00, 2020-11-01 00:00:00, 2020-12-01 00:00:00,
       2021-01-01 00:00:00],
      dtype='object')

```

andersy005: I just realized that this approach results in a `generic` index instead of a `cftimeIndex`... Let me see if there's a better way that preserves the `cftimeIndex` type.

andersy005: > I just realized that this approach results in a `generic` index instead of a `cftimeIndex`... Let me see if there's a better way that preserves the `cftimeIndex` type.


One simple workaround is to cast the results to `CFTimeIndex` after appending i.e.:

```python
In [28]: xr.coding.cftimeindex.CFTimeIndex(y)
Out[28]: 
CFTimeIndex([2020-01-01 00:00:00, 2020-02-01 00:00:00, 2020-03-01 00:00:00,
             2020-04-01 00:00:00, 2020-05-01 00:00:00, 2020-06-01 00:00:00,
             2020-07-01 00:00:00, 2020-08-01 00:00:00, 2020-09-01 00:00:00,
             2020-10-01 00:00:00, 2020-11-01 00:00:00, 2020-12-01 00:00:00,
             2021-01-01 00:00:00],
            dtype='object', length=13, calendar='gregorian', freq='MS')
```

@hCraker, let me know if this accomplishes what you want...

hCraker: This works! I will have to double check all of the test cases, but this looks good so far. Thanks

hCraker:
@andersy005 Your suggestion for `cftime_range` worked, and I have pushed the changes. Sorry about the delay. Integrating this functionality wasn't trivial. I still need to add test cases for all of the different calendar types. Would you happen to have real world data in a non-standard calendar that I could use for that? Thanks!

andersy005:
@hCraker, this is looking great so far, and I'm glad to hear that `xr.cftime_range()` worked for this use case. 

> Would you happen to have real world data in a non-standard calendar that I could use for that?

If you head over to this ftp link, you will find some sample datasets that may be useful:

ftp://ftp.cgd.ucar.edu/archive/aletheia-data/cesm-data/ocn/

Also, there are some CESM Large Ensemble datasets in the AWS: https://ncar.github.io/cesm-lens-aws/#data-catalog

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,136 @@ def month_to_season(
...
     return means.sel(
         {time_coord_name: means[time_coord_name].dt.month == months[1]})
+
+
+def calendar_average(
+        dset: typing.Union[xr.DataArray, xr.Dataset],
+        freq: str,
+        time_dim: str = None,
+        climatology: bool = False,
+        calendar: str = 'standard') -> typing.Union[xr.DataArray, xr.Dataset]:
```
andersy005, line 362: Could we set this `None` by default,
```suggestion
        calendar: str = None) -> typing.Union[xr.DataArray, xr.Dataset]:
```

and then retrieve it from the `attrs` or `encoding` of the input dataset?

```python
In [15]: ds.time.encoding
Out[15]: 
{'source': '/Users/abanihi/Library/Caches/xarray_tutorial_data/eee06791cc19e59f074155a82a7ffe90-rasm.nc',
 'original_shape': (36,),
 'dtype': dtype('float64'),
 'units': 'days since 0001-01-01',
 'calendar': 'noleap'}

In [16]: ds.time.attrs
Out[16]: {'long_name': 'time', 'type_preferred': 'int'}
```

dcherian: no, you should figure it out from `ds.time[0]` (see `infer_calendar_name` here: https://github.com/pydata/xarray/pull/5233/files#diff-e1c5f34f91e5b8a5242d9499ca9ff2765f3d5b7956c28196f80e4c399b29b1a6L325-R362) `attrs` and `encoding` record how the information was stored on disk. It could be totally out-of-date (unlikely but possible).

hCraker: @andersy005 I agree that the calendar should be retrieved from the data, but if I do this, would there be a need to keep the `calendar` kwarg? It seems redundant.

@dcherian `infer_calendar_name` returns the correct strings, so I can definitely use this approach.

hCraker: @andersy005 I have added your suggestion and updated the doc string. Please let me know if it needs to be changed.

hCraker:
Hi @andersy005 and @dcherian. I know the you two aren't formally in GeoCAT, but if you would like to review this PR when you have time, your suggestions are welcome and appreciated.

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,140 @@ def month_to_season(
...
     return means.sel(
         {time_coord_name: means[time_coord_name].dt.month == months[1]})
+
+
+def calendar_average(
+        dset: typing.Union[xr.DataArray, xr.Dataset],
+        freq: str,
+        time_dim: str = None,
+        climatology: bool = False,
+        calendar: str = None) -> typing.Union[xr.DataArray, xr.Dataset]:
```
dcherian, line 362:
```suggestion
        calendar: str = None) -> typing.Union[xr.DataArray, xr.Dataset]:
```

I would delete the calendar kwarg. it seems useless. What happens if a user provides a calendar that is not the input calendar?

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,140 @@ def month_to_season(
...
+        'year': (None, 'YS', '6MS')
+    }
+
+    try:
+        freq_dict[freq]
+    except KeyError:
+        raise KeyError(
+            f"contributed: calendar_average: bad period: PERIOD = {freq}. Valid periods include: {list(freq_dict.keys())}"
+        )
+
```
dcherian, line 420:
```suggestion
    if freq not in freq_dict:
    	raise KeyError(
            f"Received bad period {freq!r}. Expected one of {list(freq_dict.keys())!r}"
        )

```

This is the canonical python pattern. I reworded the message to be clearer

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,140 @@ def month_to_season(
...
+    # If freq is 'season', key is set to monthly in order to calculate monthly
+    # averages which are then used to calculate seasonal averages
+    key = 'month' if freq in {'season', 'year'} else freq
+
+    (format, frequency, offset) = freq_dict[key]
```
dcherian, line 425:
```suggestion
    format, frequency, offset = freq_dict[key]
```

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,140 @@ def month_to_season(
...
+
+    time_dim = _get_time_coordinate_info(dset, time_dim)
+
+    # Retrieve calendar name
+    if calendar is None:
+        calendar = xr.coding.times.infer_calendar_name(dset[time_dim])
```
dcherian, line 431: This is an internal function. I would copy it over until `.dt.calendar` is available

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,140 @@ def month_to_season(
...
+
+    if freq == 'year' and climatology:
+        climatology = False
+        warnings.warn(
+            'Cannot compute yearly climatology data since climatologies are averaged over years. To remove warning, set `climatology` to False'
```
dcherian, line 436: why warn and not error?

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,140 @@ def month_to_season(
...
+        )
+    # Average data across years
+    if climatology:
```
dcherian, line 439: I would refactor this out to a function that does the climatology. There is not much shared code; so they're conceptually different enough to be different function.

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,140 @@ def month_to_season(
...
+
+    Returns
+    -------
+    computed_dset: same type as dset
+        The computed data
+
+    Notes
+    -----
+    Seasonal averages are weighted based on the number of days in each month.
+    This means that the given data must be monotonic (i.e. data every 6 hours,
```
dcherian, line 400: you could check this with `xr.infer_freq`. If it fails, then it's possible the data is not uniformly spaced. Also Non-uniformly spaced data can still be monotonic.

dcherian (commented):
nice work.

I have a few suggestions but I ran out of time so I haven't looked through it all

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,140 @@ def month_to_season(
...
+                frequency,
+                calendar,
+                start=f'{median_yr:.0f}-{start_time}',
+                end=f'{median_yr:.0f}-{end_time}')
+
+    # Average data for each period considering the year of the period
+    else:
+        # Resample data using given frequency which preserves the year of the data
+        dset = dset.resample({time_dim: frequency}).mean().dropna(time_dim)
+        if freq == 'season' or freq == 'year':
```
andersy005, line 477:
```suggestion
        if freq in ['season', 'year']:
```

src/geocat/comp/climatology.py
```diff
@@ -322,3 +352,140 @@ def month_to_season(
...
+            key = freq
+            (format, frequency, offset) = freq_dict[key]
```
andersy005, line 479: `offset` isn't used after this line. 

```suggestion
            format, frequency, _ = freq_dict[key]
```

src/geocat/comp/climatology.py
andersy005: > I would delete the calendar kwarg


👍🏽 I concur...

src/geocat/comp/climatology.py
hCraker: I wasn't aware of this unspoken standard for error messages. Thanks!

src/geocat/comp/climatology.py
hCraker: I wasn't aware since it didn't haven an underscore in front of the name (that's the standard I have seen for internal functions). I can copy it into the file. How long do you think it will be until `.dt.calendar` will be ready?

src/geocat/comp/climatology.py
hCraker: It is a warning because the code computes the alternative instead, which is the sequential average. This tells the user that something was wrong with the input, but still outputs what the user likely wanted anyway. If I refactor the `if climatology` block out to a different function. I can change this to an error.

src/geocat/comp/climatology.py
hCraker: I had debated doing this. I thought it would be useful to keep these functionalities together, but you know the user base better than I do. If it would be less confusing for them to have these as separate functions, I can separate them.

src/geocat/comp/climatology.py
dcherian: For some reason, most pydata packages don't use that but instead restrict public functions to those documented in the documentation. https://xarray.pydata.org/en/stable/getting-started-guide/faq.html#what-parts-of-xarray-are-considered-public-api

Things in `core.*` are internal

src/geocat/comp/climatology.py
dcherian: I feel like code should not compute things the user did not ask for. If they decide they want the non-climatological output they can ask for it explicitly

src/geocat/comp/climatology.py
dcherian: SOrry that wasn't clear. The standard is the `if freq  not in freq_dict` check as opposed to checking for a `KeyError`. I just like that format of error message. It's very clear!

src/geocat/comp/climatology.py
hCraker: That is true. I believe that this suggestion will be nullified once I split the climatology and sequential calculations into two different functions as the warning/error will no longer be needed.

src/geocat/comp/climatology.py
hCraker: This has been removed in the latest commit.

src/geocat/comp/climatology.py
hCraker: `offset` is no longer used, and I had forgotten to remove it from this line and from the dictionary. It has been fixed in the latest commit along with adding your formatting changes.

src/geocat/comp/climatology.py
hCraker: I have copied over the function and left a comment saying where I got it from. Thanks for the help.

src/geocat/comp/climatology.py
hCraker: Could you clarify the difference between uniformly spaces and monotonic? I thought they were synonymous.

dcherian: as far as i know, monotonic refers to ordering so either monotonic increasing or monotonic decreasing. It says nothing of spacing between points: https://en.wikipedia.org/wiki/Monotonic_function

src/geocat/comp/climatology.py
```diff
@@ -61,6 +64,47 @@ def _setup_clim_anom_input(dset, freq, time_coord_name):
...
+    """Helper function to determine the time bounds based on the given dataset
+    and frequency and then calculate the averages of them.
+
+    Returns the dataset with the time coordinate changed to the center
+    of the time bounds.
+    """
+    if sum(x is not None for x in [start, end]) == 1:
+        raise ValueError(
+            "Both `start` and `end` must be specified or both left unspecified."
+        )
```
andersy005, line 82: A test for this would be useful...

src/geocat/comp/climatology.py
```diff
@@ -322,3 +366,190 @@ def month_to_season(
...
+
+    format, frequency = freq_dict[key]
+
+    # If time_dim is None, infer time dimension name. Confirm dset[time_dim] contain datetimes
+    time_dim = _get_time_coordinate_info(dset, time_dim)
+
+    # Check if data is uniformly spaced
+    if xr.infer_freq(dset[time_dim]) is None:
+        raise ValueError(
+            f"Data needs to be uniformly spaced in the {time_dim!r} dimension.")
```
andersy005, line 434: Some unit test for this? :)

src/geocat/comp/climatology.py
```diff
@@ -322,3 +366,190 @@ def month_to_season(
...
+        'hour': ('%m-%d %H', 'H'),
+        'day': ('%m-%d', 'D'),
+        'month': ('%m', 'MS'),
+        'season': (None, 'QS-DEC')
+    }
+
+    if freq not in freq_dict:
+        raise KeyError(
+            f"Received bad period {freq!r}. Expected one of {list(freq_dict.keys())!r}"
+        )
```
andersy005, line 506: Same here...

src/geocat/comp/climatology.py
hCraker: That unit test was recently added. It's in test_climatology.py

andersy005: For some reason, codecov is reporting this line as a miss: https://codecov.io/gh/NCAR/geocat-comp/src/d5abf855294e588403949a1bf8bc94c7cfb3f155/src/geocat/comp/climatology.py

src/geocat/comp/climatology.py
hCraker: This error is in an internal function that isn't in the name space so users can't access it. I'm actually going to remove this error altogether since it isn't really needed. I can restructure this a bit to utilize the errors that `xr.cftime_range` throws and add internal API documentation to help future developers.

src/geocat/comp/climatology.py
hCraker: Addressed in the latest PR. There is a test for this code block in both `calendar_average` and `climatology_average`

src/geocat/comp/climatology.py
hCraker: I think that was because I had a unit test for `calendar_average` but I didn't create a similar test for `climatology_average` after I split this function into two like @dcherian suggested. It should be fixed now

pilotchute (approved):
Thanks Heather!
The tests look great. 
The implementation is clear.
Approved. :)

andersy005 (approved):
Glad to see brought to completion, @hCraker 🎉

PR #161, opened 2021-06-14
Rewrite S@S functions to be compatible with dask lazy compute

anissa111:
Addresses #159

Functions to be addressed:

- [x] relhum
- [x] dewtemp
- [x] heat_index
- [x] psychrometric_constant
- [x] saturation_vapor_pressure
- [x] actual_saturation_vapor_pressure
- [x] saturation_vapor_pressure_slope

### DIFF
--- a/src/geocat/comp/crop.py
+++ b/src/geocat/comp/crop.py
@@ -139,7 +139,7 @@ def psychrometric_constant(pressure):
         pressure = np.asarray(pressure)
 
     # psychrometric constant calculation
-    # Note: no additional functions needed for type compatibility
+    # Note: no additional functions needed for xarray compatibility
     con = 0.66474e-3
     psy_const = con * pressure
 
@@ -210,9 +210,7 @@ def saturation_vapor_pressure(temperature, tfill=np.NAN):
         svp.attrs['info'] = "FAO 56; EQN 11; saturation_vapor_pressure"
 
     else:
-        # if not xarray, make sure in numpy for calculation
-        if in_type is not xr.DataArray:
-            temperature = np.asarray(temperature)
+        temperature = np.asarray(temperature)
 
         temp_c = (temperature - 32) * 5 / 9
         svp = np.where(temp_c > 0, 0.6108 * np.exp(
@@ -324,9 +322,7 @@ def saturation_vapor_pressure_slope(temperature, tfill=np.NAN):
             'info'] = "FAO 56; EQN 13; saturation_vapor_pressure_slope"
 
     else:
-        # if not xarray, make sure in numpy for calculation
-        if in_type is not xr.DataArray:
-            temperature = np.asarray(temperature)
+        temperature = np.asarray(temperature)
 
         # convert to Celsius
         temp_c = (temperature - 32) * 5 / 9
--- a/src/geocat/comp/crop.py
+++ b/src/geocat/comp/crop.py
@@ -257,28 +257,16 @@ def actual_saturation_vapor_pressure(tdew, tfill=np.NAN):
     array([1.22796262, 1.76730647, 2.50402976])
     """
 
-    x_out = False
-    if isinstance(tdew, xr.DataArray):
-        x_out = True
-        save_dims = tdew.dims
-        save_coords = tdew.coords
-
-    # convert inputs to numpy arrays if necessary
-    if not _is_duck_array(tdew):
-        tdew = np.asarray(tdew)
+    in_type = type(tdew)
 
     asvp = saturation_vapor_pressure(tdew, tfill)
 
-    # reformat output for xarray if necessary
-    if x_out:
-        heatindex = xr.DataArray(asvp, coords=save_coords, dims=save_dims)
-        heatindex.attrs[
-            'long_name'] = "actual saturation vapor pressure via Tdew"
-        heatindex.attrs['units'] = "kPa"
-        heatindex.attrs[
-            'url'] = "https://www.fao.org/docrep/X0490E/x0490e07.htm"
-        heatindex.attrs[
-            'info'] = "FAO 56; EQN 14; actual_saturation_vapor_pressure"
+    # reformat metadata for xarray
+    if in_type is xr.DataArray:
+        asvp.attrs['long_name'] = "actual saturation vapor pressure via Tdew"
+        asvp.attrs['units'] = "kPa"
+        asvp.attrs['url'] = "https://www.fao.org/docrep/X0490E/x0490e07.htm"
+        asvp.attrs['info'] = "FAO 56; EQN 14; actual_saturation_vapor_pressure"
 
     return asvp
 
--- a/src/geocat/comp/crop.py
+++ b/src/geocat/comp/crop.py
@@ -192,28 +192,31 @@ def saturation_vapor_pressure(temperature, tfill=np.NAN):
     array([1.22796262, 1.76730647, 2.50402976])
     """
 
-    x_out = False
-    if isinstance(temperature, xr.DataArray):
-        x_out = True
-        save_dims = temperature.dims
-        save_coords = temperature.coords
+    in_type = type(temperature)
 
-    # convert inputs to numpy arrays for function call if necessary
-    if not _is_duck_array(temperature):
-        temperature = np.asarray(temperature, dtype='float32')
+    if in_type is xr.DataArray:
 
-    temp_c = (temperature - 32) * 5 / 9
-    svp = np.where(temp_c > 0, 0.6108 * np.exp(
-        (17.27 * temp_c) / (temp_c + 237.3)), tfill)
+        # convert temperature to Celsius
+        temp_c = (temperature - 32) * 5 / 9
 
-    # reformat output for xarray if necessary
-    if x_out:
-        heatindex = xr.DataArray(svp, coords=save_coords, dims=save_dims)
-        heatindex.attrs['long_name'] = "saturation vapor pressure"
-        heatindex.attrs['units'] = "kPa"
-        heatindex.attrs[
-            'url'] = "https://www.fao.org/docrep/X0490E/x0490e07.htm"
-        heatindex.attrs['info'] = "FAO 56; EQN 11; saturation_vapor_pressure"
+        # calculate svp
+        svp = xr.where(temp_c > 0, 0.6108 * np.exp(
+            (17.27 * temp_c) / (temp_c + 237.3)), tfill)
+
+        # add relevant metadata
+        svp.attrs['long_name'] = "saturation vapor pressure"
+        svp.attrs['units'] = "kPa"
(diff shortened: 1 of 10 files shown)


src/geocat/comp/crop.py
```diff
@@ -196,28 +192,31 @@ def saturation_vapor_pressure(temperature, tfill=np.NAN):
...
+
+        # add relevant metadata
+        svp.attrs['long_name'] = "saturation vapor pressure"
+        svp.attrs['units'] = "kPa"
+        svp.attrs['url'] = "https://www.fao.org/docrep/X0490E/x0490e07.htm"
+        svp.attrs['info'] = "FAO 56; EQN 11; saturation_vapor_pressure"
+
+    else:
+        # if not xarray, make sure in numpy for calculation
+        if in_type is not xr.DataArray:
```
erogluorhan, line 214: This `if` does not seem to be needed (i.e. `else` above does the same check)

src/geocat/comp/crop.py
```diff
@@ -132,28 +132,24 @@ def psychrometric_constant(pressure):
...
+    # if output is xarray, add relevant metadata
+    if in_type is xr.DataArray:
+        psy_const.attrs['long_name'] = "psychrometric constant"
+        psy_const.attrs['units'] = "kPa/C"
+        psy_const.attrs[
             'url'] = "https://www.fao.org/docrep/X0490E/x0490e07.htm"
-        heatindex.attrs['info'] = "FAO 56; EQN 8; psychrometric_constant"
+        psy_const.attrs['info'] = "FAO 56; EQN 8; psychrometric_constant"
 
     return psy_const
```
erogluorhan, line 154: Even if input is of type `Array` or `float`, we are returning `np.ndarray`. We discussed returning whatever he input type was, but we are still not implementing that, right?

erogluorhan (commented):
I made some inline comments in a few functions, which may or may not be applicable to the rest of the changes in this PR. However, it may be better you first address this first review in either way, and then I am happy to re-review.

src/geocat/comp/crop.py
```diff
@@ -315,31 +302,39 @@ def saturation_vapor_pressure_slope(temperature, tfill=np.NAN):
...
+        svp_slope.attrs['units'] = "kPa/C"
+        svp_slope.attrs[
             'url'] = "https://www.fao.org/docrep/X0490E/x0490e07.htm"
-        heatindex.attrs[
+        svp_slope.attrs[
             'info'] = "FAO 56; EQN 13; saturation_vapor_pressure_slope"
 
+    else:
+        # if not xarray, make sure in numpy for calculation
+        if in_type is not xr.DataArray:
```
erogluorhan, line 328: No need for this `if`, too, as in above

src/geocat/comp/crop.py
anissa111: Correct, that is not yet implemented here

src/geocat/comp/crop.py
```diff
@@ -132,28 +132,24 @@ def psychrometric_constant(pressure):
...
     psy_const = con * pressure
 
-    # reformat output for xarray if necessary
-    if x_out:
-        heatindex = xr.DataArray(psy_const, coords=save_coords, dims=save_dims)
-        heatindex.attrs['long_name'] = "psychrometric constant"
-        heatindex.attrs['units'] = "kPa/C"
-        heatindex.attrs[
+    # if output is xarray, add relevant metadata
+    if in_type is xr.DataArray:
```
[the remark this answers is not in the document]
anissa111: No, because the `pressure = np.asarray(pressure)` needs to occur before `psy_const = con * pressure`, which I realize is confusing with my note about compatibility. I've updated the note to be about xarray compatibility.

src/geocat/comp/crop.py
```diff
@@ -132,28 +132,24 @@ def psychrometric_constant(pressure):
...
 
-    # convert inputs to numpy arrays if necessary
-    if not _is_duck_array(pressure):
+    # if not xarray, make sure in numpy for calculation
+    if in_type is not xr.DataArray:
         pressure = np.asarray(pressure)
 
+    # psychrometric constant calculation
+    # Note: no additional functions needed for type compatibility
     con = 0.66474e-3
```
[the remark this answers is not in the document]
anissa111: Possibly, but some conversion stuff needs to be done before the actual calculation where this magic number is used. See the comment below.

src/geocat/comp/crop.py
```diff
@@ -258,28 +255,16 @@ def actual_saturation_vapor_pressure(tdew, tfill=np.NAN):
...
-        heatindex.attrs[
-            'url'] = "https://www.fao.org/docrep/X0490E/x0490e07.htm"
-        heatindex.attrs[
-            'info'] = "FAO 56; EQN 14; actual_saturation_vapor_pressure"
+    # reformat metadata for xarray
+    if in_type is xr.DataArray:
+        asvp.attrs['long_name'] = "actual saturation vapor pressure via Tdew"
+        asvp.attrs['units'] = "kPa"
+        asvp.attrs['url'] = "https://www.fao.org/docrep/X0490E/x0490e07.htm"
+        asvp.attrs['info'] = "FAO 56; EQN 14; actual_saturation_vapor_pressure"
```
andersy005, line 267: This pattern is repeated throughout this module, and I am wondering if we could unify it under one function... This function would take in two inputs (the dataarray, and a dictionary of attributes) and it would return the dataarrray with updated attributes...

Something along these lines:

```python

def _update_xr_attrs(obj, attrs):
    obj.attrs.update(attrs)
    return obj 

svp_slope = _update_xr_attrs(svp_slope, attrs={'long_name': "slope saturation vapor pressure curve", 'units': "kPa/C", 
                                           'url': "https://www.fao.org/docrep/X0490E/x0490e07.htm", 
                                           'info': "FAO 56; EQN 13; saturation_vapor_pressure_slope"})
        
```

@anissa111, what do you think?

src/geocat/comp/dewtemp.py
```diff
@@ -22,33 +22,37 @@ def dewtemp(temperature, relative_humidity):
...
+    if not (np.shape(x) == np.shape(inputs[0]) for x in inputs):
+        raise ValueError("dewtemp: dimensions of inputs are not the same")
+
+    # Get input types
+    in_types = [type(item) for item in inputs]
+
+    if xr.DataArray in in_types:
+
+        # check all inputs are xr.DataArray
+        if not all(x == xr.DataArray for x in in_types):
```
andersy005, line 37: Invert any/all to simplify comparisons??

```suggestion
        if any(x != xr.DataArray for x in in_types):
```

src/geocat/comp/dewtemp.py
```diff
@@ -22,33 +22,37 @@ def dewtemp(temperature, relative_humidity):
...
-                                    dims=save_dims,
-                                    attrs=save_attrs)
+    inputs = [temperature, relative_humidity]
+
+    # ensure all inputs same size
+    if not (np.shape(x) == np.shape(inputs[0]) for x in inputs):
+        raise ValueError("dewtemp: dimensions of inputs are not the same")
+
+    # Get input types
+    in_types = [type(item) for item in inputs]
```
andersy005, line 32: I'm curious... Could this `type` sanity check move into the if condition below:

```python
if any(isinstance(x, xr.DataArray) for x in inputs):
```

src/geocat/comp/crop.py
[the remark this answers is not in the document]
erogluorhan: Please see my code suggestion above for the whole function. In that code, you will be first doing the assignment:

`psy_const = con * pressure`

Based on the data type of `pressure`, `psy_const` will be of the same type. Then, in if block, if numpy, you will be doing the similar conversion for `psy_const`, too, as follows:

```
if in_type is not xr.DataArray:
      pressure = np.asarray(pressure)
      psy_const = np.asarray(psy_const)
```

I didn't give a try to actual exceution of this suggestion, but I'd expect it to work.

What do you think?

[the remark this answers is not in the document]
anissa111: I'm almost certain it won't work because of list inputs. For plain lists, you can't multiply by a constant, which is why I needed to do a numpy conversion before `psy_const = con * pressure`.

I did a short test in my python console to demonstrate what I mean:
image

src/geocat/comp/dewtemp.py
anissa111: Yes, but I'd like to keep the types saved independently in case we can eventually do more with ensuring the output is the same type as the inputs, if that makes sense.

src/geocat/comp/crop.py
anissa111: I think that yes, that does make sense, but I appreciated the clarity/simplicity that came with setting the attributes directly instead of going through a util function, especially as the util function does not actually reduce lines. @erogluorhan what do you think about doing this?

src/geocat/comp/crop.py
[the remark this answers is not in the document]
erogluorhan: Ah yes, I missed that, thanks! Below is hopefully my final code recommendation :) that looks a bit more readable to me, but fell free to ignore if you disagree:

  ```
  # Constant
  con = 0.66474e-3

  in_type = type(pressure)

  # Psychrometric constant calculation
  # if input not xarray, make sure in numpy for calculation
  if in_type is not xr.DataArray:
      psy_const = con * np.asarray(pressure)
  # else if input is xarray, add relevant metadata for xarray output
  else:
      psy_const = con * pressure
      psy_const.attrs['long_name'] = "psychrometric constant"
      psy_const.attrs['units'] = "kPa/C"
      psy_const.attrs[
          'url'] = "https://www.fao.org/docrep/X0490E/x0490e07.htm"
      psy_const.attrs['info'] = "FAO 56; EQN 8; psychrometric_constant"

  return psy_const
```

src/geocat/comp/dewtemp.py
andersy005: It makes sense. Thank you for the clarification

src/geocat/comp/crop.py
erogluorhan: Both @andersy005 's suggestion and your point of clarity make sense to me @anissa111 . So, I am fine with either way. This attribute setting behavior is not only in this module, so if/when we decide to apply Anderson's suggestion, we could/should do that for rest of the functions (maybe in a separate PR as a small refactoring work?)

src/geocat/comp/crop.py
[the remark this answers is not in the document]
anissa111: I don't have strong feelings either way, I updated with your version!

src/geocat/comp/crop.py
anissa111: Okay, for now I'd prefer to leave as is, but if we ever decide to make that util function in the future, we could absolutely make a PR and implement all the changes at once

erogluorhan (approved):
Great work!

PR #176, opened 2021-08-11
Module name refactoring

erogluorhan:
This PR does the following refactoring:

- The functions `dewtemp`, `heat_index`, and `relhum` are all placed in a module called `meteorology.py` since they are both in the meteorology category on the NCL documentation pages.

- The module `interp_hybrid_to_pressure.py `is renamed to `interpolation.py` since we will likely create other kinds of interpolation functions in the future. That way all of the interpolation calculations could be in the same module.

- The module `climatology.py `is renamed to `climatologies.py` to avoid import issues with the function of the same name.

- Relevant documentation changes are done.

- User API is changed to list the functions in the format `import geocat.comp.<module_name>.<function_name>`

- Merging or renaming of the corresponding test scripts are also done.


hCraker (approved):
Thanks for doing this!

PR #178, opened 2021-08-17
Update skewt_params

michaelavs:
Updates to use MetPy's showalter index function


src/geocat/comp/skewt_params.py
```diff
@@ -5,67 +5,6 @@
 from metpy.units import units
 
 
-def showalter_index(pressure, temperature, dewpt):
```
erogluorhan, line 8: Existing GeoCAT-comp users might have used this function through our conda releases so far; thus, it may not be a good idea to immediately remove such a function you provided in your user API. A few alternatives to handle this instead of directly removing it:

- You may put a deprecation message into this function that states its functionality will be available via get_skewt_params from now on and this function will be deprecated (i.e. actually removed) in the upcoming version releases. 
- You could still continue providing this function signature, which would simply be a function call to metpy's function.

test/test_skewt_params.py
```diff
@@ -22,15 +22,17 @@
 
 def test_showalter_index():
 
-    result = showalter_index(p, tc, tdc)
-    expected = 21.35395610431048 * units.delta_degree_Celsius
-
-    nt.assert_equal(result, expected)
+    result = mpcalc.showalter_index(p, tc, tdc)
+    expected = [14.82771069] * units.delta_degree_Celsius
```
erogluorhan, line 26: What is your finding about this calculation discrepancy?

test/test_skewt_params.py
```diff
@@ -22,15 +22,17 @@
...
+    assert_almost_equal(result, expected, 4)
 
 
 def test_get_skewt_vars():
 
     result = get_skewt_vars(p, tc, tdc, pro)
-    expected = 'Plcl= 747 Tlcl[C]= 6 Shox= 21 Pwat[cm]= 5 Cape[J]= 0'
-
+    expected = 'Plcl= 747 Tlcl[C]= 6 Shox= 15 Pwat[cm]= 5 Cape[J]= 0'
```
erogluorhan, line 33: The same question as above

test/test_skewt_params.py
michaelavs: I think it was just left over from some other test values. I did the calculation by hand and got a value of around 15 which is consistent with what the function returned. I also ran the code against skewt_2_2 from examples and got a consistent value with that data as well.

src/geocat/comp/skewt_params.py
michaelavs: I decided to go with the second option and keep the function signature and then changing the function to call MetPy's function

test/test_skewt_params.py
erogluorhan: I think I am missing some insights here. Please see my one-by-one questions for your explanation:

> I think it was just left over from some other test values. I did the calculation by hand and got a value of around 15...

`test_skewt_params.py` has only two changes in [its history](https://github.com/NCAR/geocat-comp/commits/main/test/test_skewt_params.py) (one is its initiation, the other is what we addressesd recently due to MetPy's reversioning from 1.0 to 1.1). The expected value in that first commit in the history was "21.353962321924012" and then it has had a very slight change to "21.35395610431048" after MetPy's 1.1 release. So, "21.353xxx" was what our code always generated with this test code. I can't get why you are now able to calculate about 15 by hand. Was that about 15 or 21 previously?

> I also ran the code against skewt_2_2 from examples and got a consistent value with that data as well.

I can remember that we also had some discrepencies with what our code generated and what was in skewt_2_2 example. Knowing you got a consistent value with skewt_2_2 by using MetPy's implementation, should we assume our implementation somehow had a wrong calculation, and it is now fixed with MetPy's implementation?

src/geocat/comp/skewt_params.py
```diff
@@ -7,62 +7,27 @@
 
 def showalter_index(pressure, temperature, dewpt):
     """Calculate Showalter Index from pressure temperature and 850 hPa lcl.
-
     Showalter Index derived from [Galway1956]_:
-    SI = T500 - Tp500
 
+    SI = T500 - Tp500
```
erogluorhan, line 12: `SI` is showalter index, right? Returned `shox` also stands for showalter index. How about naming them the same to have conistency between docstring and code?

test/test_skewt_params.py
```diff
@@ -12,25 +14,69 @@
...
+# Extract Showalter Index from NCL out file and convert to int
+Shox = np.round(out['Shox'])  # Use np.round to avoid rounding issues
+NCL_shox = int(Shox[0])  # Convert to int
 
-    result = showalter_index(p, tc, tdc)
-    expected = 21.35395610431048 * units.delta_degree_Celsius
 
-    nt.assert_equal(result, expected)
+def test_shox_vals():
+    """Testing for Showalter Index only. While MetPy handles all five
```
erogluorhan, line 40: How about moving the explanation of overall testing approach (i.e. testing only showalter index) of this docstring to the beginning of this file instead? Seeing `CAPE` under `test_shox_vals` might be a bit confusing.

test/test_skewt_params.py
```diff
@@ -12,25 +14,69 @@
...
+    This test is to ensure that the values of each are being read,
+    assigned, and placed correctly in get_skewt_vars.
+    """
 
+    expected = 'Plcl= 927 Tlcl[C]= 24 Shox= 3 Pwat[cm]= 5 Cape[J]= 2958'
+    result = get_skewt_vars(p, tc, tdc, pro)
     nt.assert_equal(result, expected)
+
+
+test_shox_vals()
```
erogluorhan, line 81: When we use `pytest`, we don't explicitly call these functions. Please check other test files and change this accordingly to use `pytest` with the best practices.

erogluorhan (changes requested):

erogluorhan (approved):

PR #197, opened 2022-01-28
Try mamba to fix RTD build failures

erogluorhan:
A "failed due to excessive memory" case started here similar to one we saw recently for GeoCAT-examples. 

I attempted to fix this one with mamba as well, and it seems to have worked. 

Please see RTD build [here](https://readthedocs.org/projects/geocat-comp/builds/15907422/).


anissa111 (approved):
This looks good, I'm glad it worked!

PR #203, opened 2022-03-08
Spherical_harmonics functions

pilotchute:
Initial pull request for spherical.py and test_spherical.py
- [x] four tests passing on local machine
- [x] check codecov before merging

### DIFF
--- a/src/geocat/comp/spherical.py
+++ b/src/geocat/comp/spherical.py
@@ -114,8 +114,7 @@ def recomposition(
         phi = xr.DataArray(phi, dims=phi.dims).chunk((chunk_size))
         data = xr.DataArray(data, dims=['har']).chunk((chunk_size))
 
-    results = \
-        np.sum(
+    results = np.sum(
         np.multiply(sspecial.sph_harm(m, n, theta, phi).real, data.real),
         axis=(0),
     ) + np.sum(
--- a/src/geocat/comp/spherical.py
+++ b/src/geocat/comp/spherical.py
@@ -67,14 +67,22 @@ def decomposition(
         m = xr.DataArray(m, dims=['har']).chunk((chunk_size))
         n = xr.DataArray(n, dims=['har']).chunk((chunk_size))
         scale_mul = xr.DataArray(scale_mul, dims=['har']).chunk((chunk_size))
-        scale_dat = xr.DataArray(np.multiply(data, scale),
-                                 dims=data.dims).chunk((chunk_size))
+        scale_dat = xr.DataArray(
+            np.multiply(data, scale),
+            dims=data.dims,
+        ).chunk((chunk_size))
         theta = xr.DataArray(theta, dims=data.dims).chunk((chunk_size))
         phi = xr.DataArray(phi, dims=data.dims).chunk((chunk_size))
 
-    results = np.sum(np.multiply(scale_dat, sspecial.sph_harm(m, n, theta,
-                                                              phi)),
-                     axis=(0, 1)) * scale_mul * scale_val
+    results = np.sum(
+        np.multiply(scale_dat, sspecial.sph_harm(
+            m,
+            n,
+            theta,
+            phi,
+        )),
+        axis=(0, 1),
+    ) * scale_mul * scale_val
     return results
 
 
@@ -114,7 +122,12 @@ def recomposition(
     results = np.sum(np.multiply(
         sspecial.sph_harm(m, n, theta, phi).real, data.real),
                      axis=(0)) + np.sum(np.multiply(
-                         sspecial.sph_harm(m, n, theta, phi).imag, data.imag),
+                         sspecial.sph_harm(
+                             m,
+                             n,
+                             theta,
+                             phi,
+                         ).imag, data.imag),
                                         axis=(0))
 
     return results.real
--- a/src/geocat/comp/spherical.py
+++ b/src/geocat/comp/spherical.py
@@ -40,19 +40,19 @@ def harmonic_decomposition(
         n = np.expand_dims(n, axis=(0, 1))
         theta = np.expand_dims(theta, axis=(2))
         phi = np.expand_dims(phi, axis=(2))
-        data_scaled = np.expand_dims(np.multiply(data, scale), axis=(2))
+        scale_dat = np.expand_dims(np.multiply(data, scale), axis=(2))
 
     # if xarray, set dims and chunks for broadcast in ss.sphere_harm
     if type(data) is xr.DataArray:
         m = xr.DataArray(m, dims=['har']).chunk((chunk_size))
         n = xr.DataArray(n, dims=['har']).chunk((chunk_size))
         scale_mul = xr.DataArray(scale_mul, dims=['har']).chunk((chunk_size))
-        data_scaled = xr.DataArray(np.multiply(data, scale),
-                                   dims=data.dims).chunk((chunk_size))
+        scale_dat = xr.DataArray(np.multiply(data, scale),
+                                 dims=data.dims).chunk((chunk_size))
         theta = xr.DataArray(theta, dims=data.dims).chunk((chunk_size))
         phi = xr.DataArray(phi, dims=data.dims).chunk((chunk_size))
 
-    results = np.sum(np.multiply(data_scaled, ss.sph_harm(m, n, theta, phi)),
+    results = np.sum(np.multiply(scale_dat, ss.sph_harm(m, n, theta, phi)),
                      axis=(0, 1)) * scale_mul * scale_val
     return results
 
@@ -79,6 +79,7 @@ def harmonic_recomposition(
     if type(data) is np.ndarray:
         m = np.expand_dims(m, axis=(1, 2))
         n = np.expand_dims(n, axis=(1, 2))
+        data = np.expand_dims(data, axis=(1, 2))
         theta = np.expand_dims(theta, axis=(0))
         phi = np.expand_dims(phi, axis=(0))
 
--- a/src/geocat/comp/spherical.py
+++ b/src/geocat/comp/spherical.py
@@ -135,10 +135,11 @@ def scale_voronoi(
     data_locs_3d[:, 0] = np.sin(phi_1d) * np.sin(theta_1d)
     data_locs_3d[:, 1] = np.sin(phi_1d) * np.cos(theta_1d)
     data_locs_3d[:, 2] = np.cos(phi_1d)
-    sv = sspatial.SphericalVoronoi(data_locs_3d,
-                                   radius=1.0,
-                                   center=np.array([0, 0, 0]))
-    scale = np.array(sv.calculate_areas()).reshape(theta.shape)
+    scale = np.array(
+        sspatial.SphericalVoronoi(data_locs_3d,
+                                  radius=1.0,
+                                  center=np.array([0, 0,
+                                                   0]))).reshape(theta.shape)
     if type(theta) is xr.DataArray:
         scale = xr.DataArray(scale, dims=theta.dims).chunk(chunk_size)
     return scale
--- a/src/geocat/comp/spherical.py
+++ b/src/geocat/comp/spherical.py
@@ -44,9 +44,8 @@ def harmonic_decomposition(
         m = xr.DataArray(m, dims=['har']).chunk((chunk_size))
         n = xr.DataArray(n, dims=['har']).chunk((chunk_size))
         input_data_scaled = \
-            xr.DataArray(input_data_scaled,
-                                         dims=['lat', 'lon']).chunk(
-                                             (chunk_size))
+            xr.DataArray( \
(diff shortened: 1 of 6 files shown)


src/geocat/comp/spherical.py
```diff
@@ -0,0 +1,85 @@
...
+    if harms is None and max_harm is not None:
+        harms = []
+        for n in range(max_harm + 1):
+            for m in range(n + 1):
+                harms.append([m, n])
+
+    results = []
+    scale0 = 1 / (np.sum(input_scale, axis=(0, 1)) * ss.sph_harm(0, 0, 0, 0)**2)
+    # if input_scale is xr.DataArray:
+    #     scale0 = scale0.persist()
```
pilotchute, line 33: is this needed to prevent repeat computes of `scale0`?

src/geocat/comp/spherical.py
```diff
@@ -0,0 +1,85 @@
...
+    scale1 = scale0 * 2
+    input_data_scaled = np.multiply(input_data, input_scale)
+    # if input_data is xr.DataArray:
+    #     input_data_scaled = input_data_scaled.persist()
```
pilotchute, line 37: is this needed to prevent repeat computes of `input_data_scaled`?

pilotchute:
found a performance issue that was forcing a large .compute() call. have found a fix, but it's a large change, and I'm going to change this to a draft until the bugs are gone.

pilotchute:
fixed now, and I'm much happier with the way the computation is structured.

erogluorhan:
Is this still a draft or ready for review once the build failures and stalls are fixed?

pilotchute:
It's almost ready, I found a function in scipy last night that could calculate the voronoi area of points on a sphere, so I added a helper function to take theta and phi arrays, turn them into an array of x,y,z points to input to the `scipy.spatial.SphericalVoronoi(points((n,3)).calculate_areas()` function. 
Also added tests for those. 
I think it's about ready.
I'm not having it calculate the voronoi area by default due to the performance of the scipy function being quite slow, (and I don't want to have it calcuated every harmonic run, when if it's passed in, it only has to be calculated once and passed in any number of times. 
if scipy ever gets their voronoi math to meet `n*log(n)` speed of the fastest published area calculation algs I'll reconsider, but for now scipy is using a simpler `n^2` implementation.

pilotchute:
I still need to finish the docstrings, and give people a chance to read the code and see if I missed any sanity checks, or have anything that could cause problems in pathological cases.

erogluorhan:
> I still need to finish the docstrings, and give people a chance to read the code and see if I missed any sanity checks, or have anything that could cause problems in pathological cases.

All sound good! Just ping us when it is ready.

pilotchute:
@erogluorhan should I put in an optional parameter to instruct the decomposition to also provide an array of the `[m,n]` values for the output harmonics?

src/geocat/comp/spherical.py
```diff
@@ -0,0 +1,231 @@
...
+        thus the chunks used in the largest calculation scale at chunk_size^3
+        A chunk size of 256 is reccomended. Defaults to 'auto'
+
+    Returns
+    -------
+    decomposition : :class:`numpy.ndarray`, :class:`xarray.DataArray`
+        the spherical harmonic decomposition of the input data
+    """
+
+    scale_val = 1 / (np.sum(scale, axis=(0, 1)) *
```
erogluorhan, line 56: A description of this scale var would be helpful

test/test_spherical.py
```diff
@@ -0,0 +1,85 @@
...
+test_harmonics_np = np.array(test_harmonics)
+test_harmonics_xr = xr.DataArray(test_harmonics_np, dims=['har',
+                                                          'm,n']).compute()
+test_data_np = test_data
+test_data_xr = xr.DataArray(test_data_np, dims=['lat', 'lon']).compute()
+test_results_np = np.array(test_results)
+test_results_xr = xr.DataArray(test_results_np, dims=['har']).compute()
+
+
+def test_decomposition_np():
```
erogluorhan, line 52: Not a requirement but for getting these test cases directly debuggable in PyCharm (i.e. right click in the code and debug), you may consider putting your test case funcs into a class; something like:

```
from unittest import TestCase

class Test_spherical(TestCase):
    def test_decomposition_np(self):
    def test_...
```

src/geocat/comp/spherical.py
```diff
@@ -0,0 +1,231 @@
...
+                     sspecial.sph_harm(0, 0, 0, 0)**2)
+
+    mlist = []
+    nlist = []
+    scale_mul = []
```
erogluorhan, line 61: Describing what is done for `mlist` and `nlist` in these loops would be helpful, too. The same applies to `recomposition`, too.

src/geocat/comp/spherical.py
```diff
@@ -0,0 +1,231 @@
...
+    if type(data) is np.ndarray:
+        m = np.expand_dims(m, axis=(0, 1))
+        n = np.expand_dims(n, axis=(0, 1))
+        theta = np.expand_dims(theta, axis=(2))
+        phi = np.expand_dims(phi, axis=(2))
+        scale_res = scale_mul * scale_val
+        scale_dat = np.expand_dims(np.multiply(data, scale), axis=(2))
+
+    # if xarray, set dims and chunks for broadcast in ss.sphere_harm
+    if type(data) is xr.DataArray:
```
erogluorhan, line 84: Since SupportedTypes = Union[np.ndarray, xr.DataArray] already, I'd expect and if-else block instead of two if's. The same applies to `recomposition`, too.

src/geocat/comp/spherical.py
```diff
@@ -0,0 +1,231 @@
...
+            dims=['har'],
+        ).chunk((chunk_size)) * scale_val
+        scale_dat = xr.DataArray(
+            np.multiply(data, scale),
+            dims=data.dims,
+        ).chunk((chunk_size))
+        theta = xr.DataArray(theta, dims=data.dims).chunk((chunk_size))
+        phi = xr.DataArray(phi, dims=data.dims).chunk((chunk_size))
+
+    results = np.sum(
```
erogluorhan, line 98: I am curious if using `np.sum` and `np.multiply` would be a performance-cutter for chunked Xarrays (i.e. Dask data arrays). @anissa111 , any thoughts?

erogluorhan (commented):
The implementation here looks great! I have a few inline comments, though!

src/geocat/comp/spherical.py
pilotchute: two ifs leaves open future support for additional data types. if else would be harder to expand. and there is no performance difference. 

willing to if-else if you feel strongly about it.

src/geocat/comp/spherical.py
pilotchute: my testing showed that this was the fastest implementation. (and I did a lot of testiing) they are dask chunk compatible and I an show dask localcluster utilization graphs.

pilotchute: image

pilotchute: That shows chunked test data creation, decomposition, and recomposition.

src/geocat/comp/spherical.py
pilotchute: Sure, will add some comments.

test/test_spherical.py
pilotchute: I like this, I'll add it.

src/geocat/comp/spherical.py
pilotchute: Sure thing, it's not simple, but it's easy to explain. (the math to figure it out took me a while and a few sheets of paper.)

erogluorhan:
> @erogluorhan should I put in an optional parameter to instruct the decomposition to also provide an array of the `[m,n]` values for the output harmonics?

This sounds like a good enhancement to the function, though, I don't know of any use case

src/geocat/comp/spherical.py
erogluorhan: I see. I asked this as I remembered @anissa111 had an issue in relhum/dewtemp when she first used `np` functions, then she had to switch to using dask.array.functions. 

Did you have a chance to compare this current version vs. `dask.array.sum` and `dask.array.multiply`?

src/geocat/comp/spherical.py
```diff
@@ -0,0 +1,231 @@
...
+    chunk_size: :class: `int`, Optional
+        The size of the each edge of the dask chunks if using xarray.DataArray inputs.
+        Some arrays will be 2d, and others 1d, and the final calculation operates on a 3d array.
+        thus the chunks used in the largest calculation scale at chunk_size^3
+        A chunk size of 256 is reccomended. Defaults to 'auto'
+
+    Returns
+    -------
+    decomposition : :class:`numpy.ndarray`, :class:`xarray.DataArray`
+        the spherical harmonic decomposition of the input data
```
erogluorhan, line 53: It might be helpful to put a "See also" part that explains which NCL functions can be handled with these functions here (just as we did in climatologies.py and some other places)

src/geocat/comp/spherical.py
```diff
@@ -0,0 +1,231 @@
...
+    max_harm: :class: `int`, Optional
+        The maximum harmonic value for both m and n.
+        The total of harmonics calculated is (max_harm+1)*(max_harm+2)/2
+        Defaults to 23, for 300 total harmonics.
+
+    chunk_size: :class: `int`, Optional
+        The size of the each edge of the dask chunks if using xarray.DataArray inputs.
+        Some arrays will be 2d, and others 1d, and the final calculation operates on a 3d array.
+        thus the chunks used in the largest calculation scale at chunk_size^3
+        A chunk size of 256 is reccomended. Defaults to 'auto'
```
erogluorhan, line 48: typo "recommended"

pilotchute: done

src/geocat/comp/spherical.py
pilotchute: they also don't require separate computation calls in an if/else statement to avoid breaking on numpy.

src/geocat/comp/spherical.py
pilotchute: done

test/test_spherical.py
pilotchute: done

src/geocat/comp/spherical.py
erogluorhan: No strong feelings; just wondered the intention, and it makes sense.

src/geocat/comp/spherical.py
erogluorhan: > they also don't require separate computation calls in an if/else statement to avoid breaking on numpy.

Yeah, I agree with this making the code nicer

pilotchute:
> > @erogluorhan should I put in an optional parameter to instruct the decomposition to also provide an array of the `[m,n]` values for the output harmonics?
> 
> This sounds like a good enhancement to the function, though, I don't know of any use case

I could see a case where a user might want to know the shape a particular result's harmonic. The only reason I'm wary of implementing it is I don't think it's a great idea to have a function provide incompatible output  based on an input option. 

I think I would instead prefer to have a separate function to return the set of harmonics that the end user could then compare result(n) harm(n) to get the harmonic components.

erogluorhan:
> I think I would instead prefer to have a separate function to return the set of harmonics that the end user could then compare result(n) harm(n) to get the harmonic components.

Yes, this makes sense to me!

anissa111 (commented):
This looks good! I have a couple very minor changes, but then I'll approve!

erogluorhan (approved):

erogluorhan:
I approved but noticed later that some CI tests are failing. When they are fixed, I'll approve.

test/test_spherical.py
```diff
@@ -0,0 +1,150 @@
+import scipy.special as ss
+import numpy as np
+import math as ma
+import xarray as xr
+import sys
+import unittest
+from typing import Union
```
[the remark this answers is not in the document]
pilotchute: done

pilotchute:
@erogluorhan the CI failures are due to conda not finding an environment solution in python 3.7. it isn't due to the merge, and we have discussed not supporting python 3.7 in the 2022.03.0 release. 
Should we not worry about the CI failures on 3.7?

erogluorhan:
> @erogluorhan the CI failures are due to conda not finding an environment solution in python 3.7. it isn't due to the merge, and we have discussed not supporting python 3.7 in the 2022.03.0 release. Should we not worry about the CI failures on 3.7?

Ok, no problem then! Approved!

erogluorhan (approved):
Great work!

anissa111 (approved):
Woo!

pilotchute:
@erogluorhan this will need to be force-merged with your privileges, since the 3.7 CI is a requirement for a normal merge.

PR #205, opened 2022-03-15
Fix Build tests timing out issue

erogluorhan:
Essentially fixes the build tests timing out issue through addressing "UserWarning: Passing an object to dask.array.from_array which is already a Dask collection. This can lead to unexpected behavior." in `interpolation.py`.

PR also fixes "np.float Deprecated" warnings in `meteorology.py` and `test_meteorology.py` files

### DIFF
--- a/src/geocat/comp/interpolation.py
+++ b/src/geocat/comp/interpolation.py
@@ -287,7 +287,8 @@ def interp_sigma_to_hybrid(data: xr.DataArray,
 
         for idx, (d, s) in enumerate(zip(data_stacked, sigma_stacked)):
             output[idx, :] = xr.DataArray(
-                _vertical_remap(func_interpolate, s, sig_coords, d))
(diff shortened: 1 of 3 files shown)


pilotchute (approved):
Looks good, though this wouldn't work for numpy arrays. The type hints in the function signature are for xarrays though, so I see no reason not to approve.

erogluorhan:
> Looks good, though this wouldn't work for numpy arrays. The type hints in the function signature are for xarrays though, so I see no reason not to approve.

Great point! I initially designed this function to work only with Xarray, so, I agree, we wouldn't have a problem regarding Numpy for now (I was planning to look into the function to explore if we should Numpy as well; let me know if you have any thoughts)

One more thing: I added a small side-fix to this PR (np.float Deprecations; look at my PR description) simultaneously with your approval. Please let me know if you are still fine with your approval when the current build tests succeed again

pilotchute:
Yes, the small depreciated types fix meets my approval. :)

Issue #206, opened 2022-03-15
meteorology.py Needs Deprecations Addressed


erogluorhan:
Hi @anissa111 ,

There are several of "FutureWarning: xarray.ufuncs is deprecated. Instead, use numpy ufuncs directly" warnings in `meteorology.py`. I think you'd be the best candidate to fix them as you know the internals of the functionality in the module.

anissa111:
I'll take a look at this!

PR #208, opened 2022-03-16
Use np.ufuncs instead of xr.ufuncs

anissa111:
Addresses #206

Replaces `xr.ufuncs` with `np.ufuncs` in `meteorology.py`


anissa111:
Darn it, branched not from main. Closing and trying again.

PR #211, opened 2022-03-23
March continuous release

pilotchute:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+  schedule:
+    - cron: '0 0 * * *' # Daily “At 00:00”
+  workflow_dispatch:
 
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,9 @@
 name: CI
 on:
-  push:
   pull_request:
-  workflow_dispatch:
-  # schedule:
-  #    - cron: '0 0 * * *' # Daily “At 00:00”
+  push:
+    branches:
+      - main
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,10 +1,9 @@
 name: CI
 on:
-  push:
   pull_request:
-  workflow_dispatch:
-  # schedule:
-  #    - cron: '0 0 * * *' # Daily “At 00:00”
```
erogluorhan, line 7: Instead of removing the cron job, I'd suggest to activate and start using it. It has been much useful on GeoCAT-examples to notice the broken things on main branch especially when we didn't have frequent pull requests (e.g. updates to dependencies would start breaking our repo while we don't have any pushes to main). It could always be helpful on GeoCAT-comp as well to see the problems on main branch in a daily resolution. I'd not expect any cost of executing these cron jobs automatically daily at 00:00.

.github/workflows/ci.yml
```diff
@@ -1,10 +1,9 @@
 name: CI
 on:
-  push:
   pull_request:
-  workflow_dispatch:
```
erogluorhan, line 5: I'd suggest keeping `workflow_dispatch` to be able to manually run workflows when/if ever needed. There is no cost to have this AFAIK.

erogluorhan (changes requested):
In addition to my inline comments, I'd like you to investigate a significant point and see if it can be fixed easily in this release:

Our pip installation namespace seems not to be working, i.e. if you install geocat-comp via pip, it installs automatically geocat-f2py, but you get error with the following:

from geocat.comp import <any_geocat_f2py_function_name>

.github/workflows/ci.yml
pilotchute: ok. I'll activate it.

.github/workflows/ci.yml
pilotchute: sure thing.

erogluorhan (approved):

PR #230, opened 2022-07-14
Linint2

alinejg:
interpolation.py: added interp_wrap function and helper functions _pre_interp_wrap() and _post_interp_wrap()

test_interpolation.py: added two classes of tests for interp_wrap - test_manually_calc and test_interp_larger_dataset

/test: added 4 nc files for testing purposes

internal_api, user_api, __init__.py: updated with new functions


test/test_interpolation.py
```diff
@@ -195,3 +198,219 @@ def test_interp_sigma_to_hybrid_wrong_method(self):
...
                                         p0=_p0,
                                         method="wrong_method")
+
+
+class Test_interp_manually_calc(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        try:
+            cls.test_input = xr.load_dataset("interpolation_test_input_data.nc")
```
erogluorhan, line 208: Now that we added the files to geocat-datafiles, could you please start to access them through geocat-datafiles here, wherever applicable?

src/geocat/comp/interpolation.py
```diff
@@ -305,3 +310,198 @@ def interp_sigma_to_hybrid(data: xr.DataArray,
...
+    missing values if any surrounding points contain missing values.
+
+    Parameters
+    ----------
+    data_in : :class:`xarray.DataArray` :class:'np.ndarray'
+        Data array with data to be interpolated and associated coords. If
+        it is a np array, then lat_in and lon_in must be provided. Length must
+        be coordinated with given coordinates.
+
+    data_out: :class:'xarray.DataArray'
```
erogluorhan, line 416: I can see that the signature is having `data_out` just to get coordinates from it, but it is confusing especially since we have another `data_out` as a return value as well. How about getting rid of this and mandating the use of `lon_out` and `lat_out` only unless you have a strong preference with this?

src/geocat/comp/interpolation.py
```diff
@@ -305,3 +310,198 @@ def interp_sigma_to_hybrid(data: xr.DataArray,
...
+                data_out: supported_types = None,
+                lat_in: np.ndarray = None,
+                lon_in: np.ndarray = None,
+                lon_out: np.ndarray = None,
+                lat_out: np.ndarray = None,
+                cyclic: bool = False,
+                missing_val: np.number = None,
+                method: str = "linear") -> supported_types:
+    """Multidimensional interpolation of variables. Uses xarray.interp to
+    perform linear interpolation. Will not perform extrapolation, returns
```
erogluorhan, line 406: Xarray function can extrapolate; is it intentional we are downgrading this function to not extrapolate?

erogluorhan (commented):
This is great overall, but please address a few inline comments.

src/geocat/comp/interpolation.py
alinejg: I think dlinint2 did not allow for extrapolation, I thought it was because it is usually not suggested.

According to this, https://github.com/pydata/xarray/issues/3172 we can add extrapolation by adding fill_value='extrapolate'. I think that should be fine as long as we calculate the cyclic points first. I can add this optional parameter if you think that would be best.

src/geocat/comp/interpolation.py
alinejg: I have made changes to the signature, putting the required output coords before the optional input coords and removing data_out parameter.

src/geocat/comp/interpolation.py
erogluorhan: As long as the user explicitly wants extrapolation, I think we can do that. After you make it happen, we should be good to go I believe.

erogluorhan (approved):
Great work!

PR #232, opened 2022-07-27
version bump from 2022.04.0 to 2022.07.0

marodrig:
Bump version from 2022.04.0 to 2022.07.0


meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "geocat-comp" %}
-{% set version = "2022.04.0" %}
+{% set version = "2022.06.0" %}
```
erogluorhan, line 2: Shouldn't it be 2022.07.0 for July?

marodrig: Good catch! Fixing it.

erogluorhan (approved):

PR #234, opened 2022-08-02
Weighted pearson r correlation coefficient function

### LINKED ISSUE #229: Weighted correlation funtion
A user requested functionality for weighted correlation along with weighted standard deviation, variance, sum, and mean. This issues addresses only the weighted correlation request. 

**Links to other examples, if applicable**
[Scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.pearsonr.html) has an unweighted 1D version for array like objects
[xskillscore](https://xskillscore.readthedocs.io/en/stable/api/xskillscore.pearson_r.html#xskillscore.pearson_r) has a weighted N dimensional version that only supports `xarray.Dataset` and `xarray.DataArray`

The function should support `xarray.Dataset`, `xarray.DataArray`, as well as `numpy.ndarray`. For multidimensional data, there should be a way to specify what dimension to apply the correlation along, otherwise the correlation is done across all dimensions.

hCraker:
Closes #229 
This PR wraps [`xskillscore.pearson_r`](https://xskillscore.readthedocs.io/en/stable/api/xskillscore.pearson_r.html#xskillscore.pearson_r) and additionally provides the user the ability to pass in array like objects for the parameters.

Checklist
- [x] create function
- [x] write documentation
- [x] create test file
- [x] add function to index.rst
- [x] add any new dependencies
- [x] add function to __init__.py


docs/user_api/index.rst
```diff
@@ -22,6 +22,8 @@ GeoCAT-comp native routines
 
    geocat.comp.climatologies.month_to_season
 
+   geocat.comp.correlation.pearson_r
```
erogluorhan, line 25: Should the namespace be "correlation" or something like "stats", "statistics", etc if we have a potential to expand statistics way?

hCraker: Stats makes more sense. Let me change it

anissa111 (commented):
Looks good, I just have one question. Should we include a little more on the documentation here? I know we're mostly wrapping xskillscore, but I think it might be useful to cite the xskillscore documentation once on the description and then include actual parameter descriptions in the docstring.

What do you think?

hCraker:
> Looks good, I just have one question. Should we include a little more on the documentation here? I know we're mostly wrapping xskillscore, but I think it might be useful to cite the xskillscore documentation once on the description and then include actual parameter descriptions in the docstring.
> 
> What do you think?

I didn't want to directly quote the xskillscore documentation for the arguments, since they may change things on their end. However, one can make the argument that if xskillscore changes things substantially, then tests will fail and we'll know about it. I can add more complete descriptions, especially since xskillscore hasn't had frequent updates in a long while.

anissa111 (approved):

PR #251, opened 2022-09-12
Add xskillscore to meta.yml

hCraker:
Add `xskillscore` to `meta.yml` to ensure it is installed automatically along with other dependencies. Also I add my name to the maintainer list in `meta.yml`.

PR #252, opened 2022-09-12
Remove legacy recipe file 

marodrig:
This PR removes the meta.yaml file, as we are no longer supporting NCAR distribution channel.

Related issues:
#251


anissa111:
Haha, yeah, I was wondering if that would happen. I still think we need to have some sort of discussion on how to handle this, though. 

If we're going to keep meta.yml here, we should make sure to keep it in sync with the feedstock meta.yml.

PR #256, opened 2022-09-22
Gradients

pilotchute:
Draft PR for gradients.


pilotchute:
Checks are passing.
Moving out of draft phase.

src/geocat/comp/skewt_params.py
```diff
@@ -60,7 +60,8 @@ def get_skewt_vars(p: pint.Quantity, tc: pint.Quantity, tdc: pint.Quantity,
     Returns
     -------
     joined : :class:`str`
-        A string element with the format "Plcl=<value> Tlcl[C]=<value> Shox=<value> Pwat[cm]=<value> Cape[J]=<value>" where:
```
anissa111, line 63: What's going on with the changes in this file? I think this might be supposed to be in a different PR

test/test_fourier_filters.py
```diff
@@ -6,7 +6,7 @@
 
 # Import from directory structure if coverage test, or from installed
 # packages otherwise
-if "--cov" in str(sys.argv):
+if '--cov' in str(sys.argv):
```
anissa111, line 9: Why this change here?

test/test_spherical.py
```diff
@@ -16,6 +17,12 @@
 
 
 class Test_Spherical(unittest.TestCase):
+    test_results_np = None
+    test_data_np = None
+    test_harmonics_np = None
+    phi_np = None
+    theta_np = None
+    test_scale_np = None
```
anissa111, line 25: I think this should probably be in a different PR as well. Possibly related to #263?

src/geocat/comp/comp_util.py
```diff
@@ -1,6 +1,10 @@
+import copy
+
 import numpy as np
+import xarray as xr
 
 
+# remove xarray has implemented _is_duck_array() OBE
 def _is_duck_array(value):
     """Returns True when ``value`` is array-like."""
     if isinstance(value, np.ndarray):
```
anissa111, line 10: I think we should remove the rest of the changes to this file since the new util functions were removed

src/geocat/comp/gradient.py
```diff
@@ -0,0 +1,316 @@
...
+    Parameters
+    ----------
+    data : :class:`numpy.ndarray`, :class:`xarray.DataArray`
+        n-dimensional dataset, with orthographic latitude longitude coordinates
+
+    Returns
+    -------
+    gradients : list of :class:`numpy.ndarray`, list of
+    :class:`xarray.DataArray`
+        longitudinal and latitudinal gradients calculated using th WGS84 geoid.
```
anissa111, line 134: A tiny usage example would be cool here.

Also, we should talk about converting your working notebooks into something we can add to the geocat-comp gallery after this is all merged in

test/test_gradient.py
```diff
@@ -0,0 +1,50 @@
...
+    def setUpClass(cls):
+        cls.test_data = xr.load_dataset(
+            'test/gradient_test_data.nc').to_array().squeeze()
+        cls.test_results_lon = xr.load_dataset(
+            'test/gradient_test_results_longitude.nc').to_array().squeeze()
+        cls.test_results_lat = xr.load_dataset(
+            'test/gradient_test_results_latitude.nc').to_array().squeeze()
+        cls.results = gradient(cls.test_data)
+        cls.results_lon = cls.results[0]
+        cls.results_lat = cls.results[1]
```
anissa111, line 36: Feel free to disagree with me, but I think the actual function running should be performed in the tests themselves, not during setup.

src/geocat/comp/gradient.py
```diff
@@ -0,0 +1,316 @@
...
+    integral
+    """
+    return rad_lat_wgs84(lat) * np.cos(lat * d2r) * lon * d2r
+
+
+def gradient(data: xr.DataArray) -> [xr.DataArray]:
+    r"""Extract and return the gradient values of a dataset at each point in the
+    dataset. Assuming that the data points are on the surface of the WGS84
+    ellipsoid.
+
```
anissa111, line 124: I think we could expand on what methods you're using in the function in the docstring, thoughts?

Additionally, we still need to add the warnings we've discussed about the edge removal and the lack of dask compatibility

anissa111 (commented):
I've left a few comments below. 

Also, I think we could do more with the docstrings here (apart from what I've already mentioned), but I think it will be fine to wait until the documentation changes are merged before taking a crack at that because we'll likely have to go back over it anyway.

src/geocat/comp/skewt_params.py
pilotchute: I never intentionally edited that file, I'll revert it.

test/test_fourier_filters.py
pilotchute: Was a cleanup because I was using that test file for reference and it bothered me, not needed in this pr, will remove if requested.

test/test_spherical.py
pilotchute: This prevents issues due to accessing class method prior to instantiation of the class, will move to different pr if needed.

src/geocat/comp/comp_util.py
pilotchute: sure

src/geocat/comp/gradient.py
pilotchute: Yeah notebooks are currently in geocat-scratch/akootz, will get them cleaned up and added to wherever as a second proirity while preparing geocat-viz tutorial.

test/test_gradient.py
pilotchute: fine by me. since we get both dimensions in one call, I figured it would be best to call once, then have the checks be separate, but no harm in running it twice.

src/geocat/comp/gradient.py
pilotchute: yeah the warnings about data shape changes are needed, though it is fully dask compatible.

pilotchute: just not n-dimension compatible, since it requires 2d arrays.

anissa111: Oh, great! In that case, could you add dask compatibility tests?

test/test_gradient.py
anissa111: Or combining the two tests in to one?

src/geocat/comp/gradient.py
anissa111: Yeah, understood

test/test_fourier_filters.py
anissa111: Yeah let's remove it here and re-address it in the future testing changes PRs

test/test_spherical.py
anissa111: Yeah, why don't we move this over to #263 when we get a branch going on that

src/geocat/comp/gradient.py
pilotchute: Sure should be simple.

test/test_gradient.py
pilotchute: I'd rather run the tests twice, and have two failure messages than have one failure for two possible causes.

src/geocat/comp/gradient.py
```diff
@@ -0,0 +1,316 @@
...
+        convolve(datascalelon, sumkernlon, mode='wrap', origin=0),
+        dims=('lon', 'lat'),
+    )
+
+    # do lat and lon differences, can use the gradient kernel
+
+    return [datasumlon[1:-1, 1:-1], datasumlat[1:-1, 1:-1]]
+
+
+# '''
```
erogluorhan, line 201: Please document for future why we need the below commented-out codes to be kept

src/geocat/comp/gradient.py
```diff
@@ -0,0 +1,316 @@
+from typing import Union
+
+import numpy as np
+import xarray as xr
+from scipy.ndimage import convolve
+
+SupportedTypes = Union[np.ndarray, xr.DataArray]
+XTypes = Union[xr.DataArray, xr.Dataset]
+
+d2r = 1.74532925199432957692369e-02  # degrees to radians conversion factor
```
erogluorhan, line 10: May `DEG_TO_RAD` better compatible with PEP8, not sure though?

src/geocat/comp/gradient.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+def rad_lat_wgs84(lat: SupportedTypes,):
```
erogluorhan, line 13: Is this really intended to be in the user API? (Doesn't seem so as it is not imported in __init__.py). If not, then please name accordingly with leading underscores,  remove from the user API index.rst, and instead add into internal API index.rst.

src/geocat/comp/gradient.py
```diff
@@ -0,0 +1,316 @@
...
+        3.93234660296994989005722e-23 * lat ** 12 - \
+        4.67346944715944948938249e-18 * lat ** 10 + \
+        5.23053397296821359502194e-13 * lat ** 8 - \
+        2.61499198791487361911660e-08 * lat ** 6 + \
+        6.57016735969856188450312e-04 * lat ** 4 - \
+        6.50322744547926518806010e+00 * lat ** 2 + \
+        6378137.0
+
+
+def arc_lat_wgs84(lat: SupportedTypes,):
```
erogluorhan, line 56: The same with the previous comment regarding user API.

src/geocat/comp/gradient.py
```diff
@@ -0,0 +1,316 @@
...
+        5.27941504241845042397655e-26 * lat ** 13 - \
+        7.41522084948104995730556e-21 * lat ** 11 + \
+        1.01433377184128234589268e-15 * lat ** 9 - \
+        6.52003144319804471773524e-11 * lat ** 7 + \
+        2.29342105667605006614878e-06 * lat ** 5 - \
+        3.78342436432244021373642e-02 * lat ** 3 + \
+        111319.490793273572647713 * lat
+
+
+def arc_lon_wgs84(
```
erogluorhan, line 98: Same with the previous comment

src/geocat/comp/gradient.py
```diff
@@ -0,0 +1,316 @@
...
+    and a needed step for the arc_lat_wgs84 function to avoid the elliptic
+    integral
+    """
+    return rad_lat_wgs84(lat) * np.cos(lat * d2r) * lon * d2r
+
+
+def gradient(data: xr.DataArray) -> [xr.DataArray]:
+    r"""Extract and return the gradient values of a dataset at each point in the
+    dataset. Assuming that the data points are on the surface of the WGS84
+    ellipsoid.
```
erogluorhan, line 123: Please try documenting if there are any limitations of the current implementation.

erogluorhan (commented):
Thanks for putting this together! Please address a few inline comments.

src/geocat/comp/gradient.py
pilotchute: It is intended for the user api. The values used to determine distance along the wgs84 cannot be hidden from the end users, as it may become important in later data analysis.

src/geocat/comp/gradient.py
pilotchute: The same with the previous comment regarding user API.

src/geocat/comp/gradient.py
pilotchute: Same with the previous comment

src/geocat/comp/gradient.py
pilotchute: Will do

src/geocat/comp/gradient.py
pilotchute: I have plans to come back through to improve the function a bit, to avoid having to trim the edges of the dataset, and to consider just using a numpy style gradient rather a left-right gradient average. 

I plan to keep the comment block in as a reference to the "optimal" output, that was also too slow to publish.

src/geocat/comp/gradient.py
pilotchute: It is only used once, on line 117, I'll just move the constant down there.

src/geocat/comp/gradient.py
erogluorhan: Please put a similar explanation to the beginning of the commented-out code block.

pilotchute: Yeah I had an improvement to the underlying implementation that I am content enough with that I see no reason to retain the elementwise implementation.

src/geocat/comp/gradient.py
pilotchute: dask compatibility tests added.

src/geocat/comp/gradient.py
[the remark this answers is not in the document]
pilotchute: it's been removed.

test/test_fourier_filters.py
pilotchute: done

test/test_spherical.py
pilotchute: reverted for now, will add in for a misc improvement iteration task.

src/geocat/comp/comp_util.py
pilotchute: reverted.

erogluorhan (approved):

src/geocat/comp/__init__.py
```diff
@@ -9,6 +9,7 @@
 from .fourier_filters import (fourier_band_block, fourier_band_pass,
                               fourier_filter, fourier_high_pass,
                               fourier_low_pass)
+from .gradient import gradient
```
anissa111, line 12: If the other functions are intended to be public API, they should be included here, yes?

```suggestion
from .gradient import gradient, arc_lon_wgs84, arc_lat_wgs84, rad_lat_wgs84
```

src/geocat/comp/gradient.py
```diff
@@ -0,0 +1,167 @@
...
+def rad_lat_wgs84(lat: SupportedTypes,):
+    r"""The radius calculation for the wgs84 ellipsoid at a latitude uses a
+    taylor series from.
+
+    .. math:
+        :nowrap:
+        radius = \sqrt{a^2 \cdot cos(lat)^2+b^2 \cdot sin(lat)^2}
+    This returns the radius of the ellipsoid for a given latitude
+    This is accurate to within floating point error.
+
```
anissa111, line 21: Since these are public, I think we should include the parameters and returns sections in the docstrings here and in the other functions.

anissa111 (changes requested):
I've pushed an extremely minor change to the math blocks that weren't working on remote or local.

I've also included some comments below, but then I think we'll be good to merge!

docs/user_api/index.rst
```diff
@@ -31,6 +31,19 @@ Crop
...
+.. currentmodule:: geocat.comp.gradient
+.. autosummary::
+   :nosignatures:
+   :toctree: ./generated/
+
+   geocat.comp.gradient.gradient
+   geocat.comp.gradient.arc_lat_wgs84
+   geocat.comp.gradient.arc_lon_wgs84
+   geocat.comp.gradient.rad_lon_wgs84
+   geocat.comp.interpolation.interp_hybrid_to_pressure
```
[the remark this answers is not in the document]
pilotchute: you are correct.

src/geocat/comp/gradient.py
pilotchute: done

anissa111 (approved):

PR #266, opened 2022-10-06
Module reorganization

### LINKED ISSUE #248: Reorganize function modules and categories
During the weekly geocat-comp meeting, we discussed reorganizing the modules into more cohesive categories. The discussed changes are as follows:
- [x] util: remove this module since xarray has the same functionality now
- [x] crop:merge with meteorology module
- [x] eofunc: merge with stats module
- [x] errors: consider removing since specific error messages can be easily created
- [x] polynomial: double check that these functions aren't already available in xarray
- [x] skewt_params: move `showalter_index` to meteorology module, move `get_skewt_vars` to geocat-viz
- [x] add deprecation warnings for changes that will break things

hCraker:
Closes #248
This PR merges some modules and moves some functions around for better organization. Deprecation warnings have been added and will persist for a few releases at the very least.


anissa111 (commented):
Yeah, this looks good! I just have a few questions:

1. From the documentation, it is unclear that these functions have been deprecated. I think we should look into using the deprecation decorator or using the deprecation rst block to make clear in our docs that we're deprecating this.
2. Did we also want to do a re-arrange of the function categories on the docs? If you don't think we should, that's okay with me, but if so, I think this PR would be a good place

hCraker:
> Yeah, this looks good! I just have a few questions:
> 
> 1. From the documentation, it is unclear that these functions have been deprecated. I think we should look into using the deprecation decorator or using the deprecation rst block to make clear in our docs that we're deprecating this.
> 2. Did we also want to do a re-arrange of the function categories on the docs? If you don't think we should, that's okay with me, but if so, I think this PR would be a good place

Thanks for mentioning this. I forgot to change the docs. I'll be sure to do that so the deprecation is more clear.

hCraker:
@anissa111 So the pre-commit hook doesn't like that the deprecation Sphinx directive needs the message to be indented in the doc string. Can we change the hook?

src/geocat/comp/crop.py
```diff
@@ -2,15 +2,20 @@
 import typing
 import warnings
 import xarray as xr
-from .comp_util import _is_duck_array
+
+import geocat.comp.meteorology as gcm
```
anissa111, line 6: Do we want to demonstrate this import method? I think it might be cleaner in-module to show the exact import we want if we do something like:

```python
import geocat.comp as gc

gc.meteorology.max_daylight(...)
```

This is super picky though, feel free to disagree

src/geocat/comp/crop.py
```diff
@@ -2,15 +2,20 @@
...
 
 def max_daylight(
     jday: typing.Union[np.ndarray, xr.DataArray, list,
                        float], lat: typing.Union[np.ndarray, xr.DataArray, list,
                                                  float]
 ) -> typing.Union[np.ndarray, xr.DataArray, float]:
-    """Computes maximum number of daylight hours as described in the Food and
+    r""".. deprecated:: 2022.10.0 The ``crop`` module is deprecated.
+        ``max_daylight`` has been moved to the ``meteorology`` module for
+        future use.
```
anissa111, line 16: To make it super clear, should we include the correct import method in the deprecation warning? I think I might be confused if I got that warning in another package. Something like 

```suggestion
    r""".. deprecated:: 2022.10.0 The ``crop`` module is deprecated.
        ``max_daylight`` has been moved to the ``meteorology`` module for
        future use. Use ``geocat.comp.max_daylight`` or
        ``geocat.comp.meteorology.max_daylight`` for the same functionality.
```

anissa111 (commented):
Some super minor comments below. 

Also! We should do a local install of this branch and test the imports of various namespaces to make sure we haven't broken anything and that imports from ``geocat.comp`` namespace are by default the reorganized versions because technically we have some duplication in that namespace now. Let me know if you've already tried this!

src/geocat/comp/crop.py
hCraker: This is definitely more clear. I'll go through and change that for all of the deprecation warnings

anissa111 (commented):
Okay, I've completed the namespace testing and I just have a couple comments.

- python seems to be able to figure out which function to import when we overload the namespace, but we should make sure when the package is released.
- We should move the tests to the corresponding test files for all files

src/geocat/comp/crop.py
```diff
@@ -2,15 +2,21 @@
 import typing
 import warnings
 import xarray as xr
-from .comp_util import _is_duck_array
+
+import geocat.comp as gc
```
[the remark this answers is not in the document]
hCraker: I'll work on that today. Thanks for catching this

hCraker:
@anissa111 So the tests for the deprecated functions had already been moved to their new locations, but I did forget to delete the old tests. It should be updated now

src/geocat/comp/eofunc.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 import xarray as xr
 from eofs.xarray import Eof
+import geocat.comp as gc
```
anissa111, line 7: Sorry, I should have specified the same change here from my previous review

src/geocat/comp/skewt_params.py
```diff
@@ -4,11 +4,18 @@
 import numpy as np
 import pint.quantity
 from metpy.units import units
+import warnings
+import geocat.comp as gc
```
anissa111, line 8: Same here

src/geocat/comp/skewt_params.py
```diff
@@ -34,13 +41,22 @@ def showalter_index(pressure: pint.Quantity, temperature: pint.Quantity,
...
+        "for the same functionality.", DeprecationWarning)
+    return gc.meteorology.showalter_index(pressure, temperature, dewpt)
 
 
 def get_skewt_vars(p: pint.Quantity, tc: pint.Quantity, tdc: pint.Quantity,
                    pro: pint.Quantity) -> str:
-    """This function processes the dataset values and returns a string element
+    r""".. deprecated:: 2022.10.0 The ``skewt_params`` module is deprecated, and
+        ``get_skewt_vars`` has been moved to the
+        `geocat.viz <https://geocat-viz.readthedocs.io/en/latest/index.html>`__
```
anissa111, line 56: It'd be nice to make sure this is actually released over there when we publish this

anissa111 (commented):
A couple stragglers and then pre-commits and this should be good to go!

src/geocat/comp/skewt_params.py
hCraker: Here's the PR we need to watch https://github.com/NCAR/geocat-viz/pull/90. The checks are failing because of some weirdness in GeoCAT-examples. I'm going to try and rerun them

anissa111 (approved):

PR #273, opened 2022-10-18
Make local install for docs better

### LINKED ISSUE #271: change package install method to avoid needing to repeatedly pip install from local
Currently, for docstrings to be updated from local changes, we need to redo a pip install. This is unideal and annoying.

anissa111:
Closes #271

Summary of changes:

- uses docs env file to make an editable local install
- adds a print statement to conf.py that prints the sys path at time of documentation generation for debugging purposes
- removes a sys path add that isn't needed with new pip install method

### DIFF
--- a/build_envs/docs.yml
+++ b/build_envs/docs.yml
@@ -19,3 +19,6 @@ dependencies:
     - sphinx-design
     - nbsphinx
     - xskillscore
+    - pip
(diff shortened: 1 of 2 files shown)


erogluorhan (approved):
This looks good!

pilotchute (approved):

Issue #277, opened 2022-10-27
Pint issues


anissa111:
Pint seems to have removed `Quantity` from their top level namespace in their most recent release, 0.20. 

This is breaking local installs of the package and our recent 2022.10.0 release.

I think this makes a good case for us pinning package dependencies at their most current compatible level when we make releases, too, since we released a package that was okay and then had a dependency break.

We'll need to address the pint changes, but for the moment I think I'm just going to pin pint in this repo. @marodrig, could you see if we'll need to do a pinned 2022.10.1 release?

PR #278, opened 2022-10-27
Pin pint

anissa111:
Temporary solution to #277 

Summary of changes:
- pins pint to <0.20 in `environment.yml` and `upstream-dev-environment.yml`
- removes ncar channel from above files, as is no longer relevant


erogluorhan (approved):
Thanks for this!

Issue #280, opened 2022-10-27
Dependencies updates can break releases


anissa111:
Currently, the latest version (2022.10.0) of geocat-comp breaks on import (See [GH actions logs](https://github.com/NCAR/geocat-comp/actions/runs/3333475484/jobs/5515429328#step:6:158))
```python
>>> import geocat.comp
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/anissaz/opt/miniconda3/envs/test_import/lib/python3.10/site-packages/geocat/comp/__init__.py", line 16, in <module>
    from .skewt_params import get_skewt_vars, showalter_index
  File "/Users/anissaz/opt/miniconda3/envs/test_import/lib/python3.10/site-packages/geocat/comp/skewt_params.py", line 5, in <module>
    import pint.quantity
ModuleNotFoundError: No module named 'pint.quantity'
```

PR #277 addresses this, but the published package is still broken.

We should do the following:
- Patch release of 2022.10.0 as soon as possible that includes the pint pinning
- Look in to how other packages handle pinning dependencies for releases so this doesn't happen again
- Consider making a nightly released package import test to catch potential issues as soon as possible

anissa111:
@dcherian thank you very much, I'll look in to that!

erogluorhan:
> > Consider making a nightly released package import test to catch potential issues as soon as possible
> 
> Highly recommend [xarray's infrastructure](https://github.com/pydata/xarray/blob/main/.github/workflows/upstream-dev-ci.yaml) for this. keewis just created a new action with most of the scripting (https://github.com/xarray-contrib/issue-from-pytest-log) so it should be relatively easy now

Looks like we already have this infrastructure in our [upstream workflow](https://github.com/NCAR/geocat-comp/blob/main/.github/workflows/upstream-dev-ci.yml), don't we? However, we may want to modify our [upstream environment file](https://github.com/NCAR/geocat-comp/blob/main/build_envs/upstream-dev-environment.yml) to install as many of those dependencies as we want to install from their GitHub repo. 

@dcherian xarray's workflow seems to be conda-installing everything, then removing many of them, and pip-installing through GitHub repositories. What is the intention with conda-install and uninstall stage? Couldn't they directly be pip-installed?

dcherian:
Perhaps to keep dependencies at the same versions as in the main env?

The real magic is the automated opening of an issue when the upstream-dev env fails.

[...]

PR #281, opened 2022-10-27
Pins Pint version

marodrig:
Build number increase with pinning of Pint version.

fixes:  #280


anissa111:
@marodrig, I think the plan was to just bump the build number since no functionality was added

marodrig:
@anissa111 Thank you! Just pushed a new commit with the correct version/build number.

anissa111 (approved):

PR #293, opened 2022-11-15
Update dependency list

### LINKED ISSUE #267: Should we keep listing dependencies under Installation docs?
There is a section named "Required dependencies for building and testing GeoCAT-comp" in `installation.rst`. 

It is probably helpful for the reader, but it requires the developers to always make sure the list is up-to-date with the actual dependencies that are listed in `environment.yml`, `requirements.txt`, or conda recipes. 

Should we consider reemoving that section or linking the reader to one of the above files?

hCraker:
Closes #267 
This PR updates the dependency list on the geocat-comp installation page. The contents of `environment.yml` are now printed in the documentation so they are always up to date. The dependencies needed only for testing have been denoted as such and moved to the bottom of the dependency list.

This is what that section of the installation instruction looks like now.
image


anissa111:
I think this is a good way to handle this. I have a few thoughts:
- I think we should add more explanatory detail to the environment file, for example, that the `pre-commit` package is a dependency for development and building, but not running the package, you know?
- Should we include the docs environment file in another section as well?
- Should we explain the pinnings we have at this time and why?

hCraker:
> * I think we should add more explanatory detail to the environment file, for example, that the `pre-commit` package is a dependency for development and building, but not running the package, you know?

Good point. I can change that.

> * Should we include the docs environment file in another section as well?

I don't see why since our average user wouldn't need to build the docs, and contributors can look at the contributors guide for more info on building them. I don't think this info is needed on the website, but we can add a comment to `docs.yml` mentioning the contributor guide instructions.

> * Should we explain the pinnings we have at this time and why?

I'd love to do this, but we'd run into the same issue of the `installation.rst` page being out of date. We could include explanations in `environment.yml` since that will be kept up to date.

anissa111:
> I don't see why since our average user wouldn't need to build the docs, and contributors can look at the contributors guide for more info on building them. I don't think this info is needed on the website, but we can add a comment to docs.yml mentioning the contributor guide instructions.

I agree, but I supposed I was viewing this section as more aimed towards more involved users or potentially developers/contributors. When we get around to it, maybe this section should be moved to a future "For Developers" section and include the documentation requirements, too.

> I'd love to do this, but we'd run into the same issue of the installation.rst page being out of date. We could include explanations in environment.yml since that will be kept up to date.

Yeah! I think we might be able to include extremely brief explanation for pinning in the environment file. It might not be conventional exactly, but I think it would help us (and users potentially).

hCraker:
@anissa111 

> I agree, but I supposed I was viewing this section as more aimed towards more involved users or potentially developers/contributors. When we get around to it, maybe this section should be moved to a future "For Developers" section and include the documentation requirements, too.

I can do that in this PR. If we label is as "For Developers" the hopefully less experienced users won't try to build the docs and run into a lot of issues.

> Yeah! I think we might be able to include extremely brief explanation for pinning in the environment file. It might not be conventional exactly, but I think it would help us (and users potentially).

I don't remember the exact reasons we pinned things. Would you be able to make a commit with that information or type it in a comment?

hCraker:
@anissa111 Great! I just need an approval for this PR and we are good to merge.

PR #304, opened 2022-11-21
Add check to avoid calculating monthly averages on monthly data

hCraker:
This PR fixes a bug found in `climatology_average`. When monthly data is passed into the function to compute seasonal averages, the program attempts to compute monthly averages from monthly data which throws an error. A small check has been added to make sure the data isn't monthly before computing the monthly averages and then seasonal averages.


anissa111:
@hCraker, what was the status of this? I know we talked offline but remind me, are you waiting on my review here for this as-is?

anissa111 (approved):

PR #312, opened 2022-12-07
CI failures quick patch

anissa111:
Resolves #308 
Related: https://github.com/pydata/xarray/issues/7362

This PR removes `xr.set_options(keep_attrs=True)` from `climatology.py`. This was causing the bug described in the xarray issue above. This change will clear up the CI failures and will need to be followed by another RP that incorporates an optional argument of `keep_attrs=True` for the climatology functions affected by the change.

PR #315, opened 2022-12-07
Add `keep_attrs` to functions

### LINKED ISSUE #314: Add `keep_attrs` kwarg to all `climatology.py` functions
Changes in xarray have resulted in an error when passing numpy objects into some `meteorology.py` functions. It has been determined in https://github.com/NCAR/geocat-comp/pull/312 that this arose from the global xarray option `keep_attrs` being set to `True` at the beginning of `climatology.py`. That line is being removed in #312, but to maintain the same the same behavior of `climatology.py` functions, a new kwarg must be added to each function in the file. This issue addresses that.

hCraker:
Closes #314

I cannot test these functions with `np.ndarray` inputs since they don't yet explicitly support that data structure. As far as I can tell, the aggregation functions for `xarray.core.groupby` and `xarray.core.resample` objects are the only ones used in this file that have `keep_attrs` as an argument.

TODO from our discussion below:
- [x] set the default to None to preserve xarray default behavior
- [x] add a test for the new kwarg

### DIFF
--- a/src/geocat/comp/climatologies.py
+++ b/src/geocat/comp/climatologies.py
@@ -229,7 +229,10 @@ def climatology(
     # TODO: Compute weighted climatologies when `time_bounds` are available
     clim = grouped.mean(time_coord_name, keep_attrs=keep_attrs)
     if time_invariant_vars:
-        return xr.concat([dset[time_invariant_vars], clim], dim=time_coord_name)
+        if keep_attrs == False:
+            return xr.concat([dset[time_invariant_vars], clim], combine_attrs='drop', dim=time_coord_name)
+        else:
+            return xr.concat([dset[time_invariant_vars], clim], dim=time_coord_name)
     else:
         return clim
 
--- a/src/geocat/comp/climatologies.py
+++ b/src/geocat/comp/climatologies.py
@@ -230,9 +230,12 @@ def climatology(
     clim = grouped.mean(time_coord_name, keep_attrs=keep_attrs)
     if time_invariant_vars:
         if keep_attrs == False:
-            return xr.concat([dset[time_invariant_vars], clim], combine_attrs='drop', dim=time_coord_name)
+            return xr.concat([dset[time_invariant_vars], clim],
+                             combine_attrs='drop',
+                             dim=time_coord_name)
         else:
-            return xr.concat([dset[time_invariant_vars], clim], dim=time_coord_name)
+            return xr.concat([dset[time_invariant_vars], clim],
+                             dim=time_coord_name)
     else:
         return clim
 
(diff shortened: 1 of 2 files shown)


hCraker:
@anissa111 I've gotten the checks to pass. I think this is good to go. Can you review it when you have a chance?

src/geocat/comp/climatologies.py
```diff
@@ -60,12 +60,13 @@ def _setup_clim_anom_input(dset, freq, time_coord_name):
...
-                        cftime.datetime], end: typing.Union[str,
-                                                            cftime.datetime]
-) -> typing.Union[xr.Dataset, xr.DataArray]:
+        dset: typing.Union[xr.Dataset, xr.DataArray],
+        time_dim: str,
+        freq: str,
+        calendar: str,
+        start: typing.Union[str, cftime.datetime],
+        end: typing.Union[str, cftime.datetime],
+        keep_attrs: bool = False) -> typing.Union[xr.Dataset, xr.DataArray]:
```
dcherian, line 69: I suggest a default of `None` to preserve Xarray's default and allow users to configure using `xr.set_options`.

Actually I'm not sure you want a new kwarg. An example illustrating 
``` python
with xr.set_options(keep_attrs=False):
    # do climatology
```
might be enough?

anissa111: @dcherian This PR is choosing to set the new kwarg to true it to preserve the previous climatology functionality where `xr.set_options(keep_attrs=True)` was set at the top of the `climatology.py` file. 

I see your point here, though, about setting it to `None` to preserve default xarray functionality. This might break current users' workflow slightly, but perhaps it's better to do that now rather than later.

Given that the main goal of the PR is to preserve the `keep_attrs=True` in most places of `climatology.py`, I think I would lean towards adding the optional kwarg instead of something like `with xr.set_options(keep_attrs=True):`, could you explain what advantages that would have over the optional kwarg?

hCraker: Bumping this since AMS is over @dcherian @anissa111

dcherian: > This PR is choosing to set the new kwarg to true

They seem to be all False in this diff.

> could you explain what advantages that would have over the optional kwarg?

just that you have less code to maintain/test. This is pretty minor. I don't have a strong opinion.

hCraker: > They seem to be all false in this diff.

After retracing my steps, you're right that the default should be True to maintain the previous behavior (see https://github.com/NCAR/geocat-comp/pull/312). But you bring up a good point of setting the default to None to preserve the xarray defaults. I don't think this will break users' code that much since it's a meta data thing and not related to the calculations themselves. I'll change it to None now.

> just that you have less code to maintain/test.

I think it's worth the effort to allow users to specify whether or not they want to keep attributes, especially since it should only add one extra test. Thanks for making us aware or other options and contributing to this discussion!

anissa111: > I think it's worth the effort to allow users to specify whether or not they want to keep attributes, especially since it should only add one extra test.

I'm inclined to here, I think the method implemented in this PR allows for a tiny bit more control.

@hCraker this PR should add that extra test, though.

hCraker:
Tests have been added and they revealed some unexpected behavior. I modified the functions to preserve or drop attributes as expected.

anissa111 (approved):

PR #319, opened 2022-12-14
Fix pint bug in showalter_index and deprecate

### LINKED ISSUE #305: Upstream CI: Pint
In the most recent releases of geocat-comp, we've opted for pinning pint to >0.20, but we'll need to update our codebase to reflect the most recent update.

See [failing upstream CI tests](https://github.com/NCAR/geocat-comp/actions/runs/3560640392/jobs/5980811808#step:6:152)

hCraker:
Closes #305 
This PR patches a bug with `meteorology.showalter_index` that was causing issues with our CI. In addition, the function has been deprecated in favor of directly using `metpy.calc.showalter_index`.

### DIFF
--- a/build_envs/docs.yml
+++ b/build_envs/docs.yml
@@ -9,7 +9,7 @@ dependencies:
     - geocat-viz
     - xarray
     - netcdf4
-    - pint<0.20
+    - pint
     - ipykernel
     - ipython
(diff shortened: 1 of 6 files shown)


anissa111 (approved):
After the pre-commits pass looks good!

PR #332, opened 2023-01-25
Replace and deprecate `anomaly`

### LINKED ISSUE #265: Replace and deprecate `climatology` and `anomaly`
The functions `climatology` and `anomaly` don't appropriately weight monthly data by the number of days in each month. The functions `climatology_average` and `calendar_average` effectively replace `climatology`, but `anomaly` still relies on the inaccurate `climatology` function.

- [x] replace or fix `anomaly`
- [x] add deprecation warning for `climatology`
- [x] add deprecation warning for `anomaly` if necessary

hCraker:
Closes #265
This PR adds a new function called `climate_anomaly` which is meant to replace the old `anomaly` function. This replacement is happening because `anomaly` is based off of the older `climatology` which does not correctly weight monthly data. `climate_anomaly` relies on `climatology_average` and `calendar_average` which both correctly weight monthly data. `climate_anomaly` also has xarray attribute handling which `anomaly` did not have.

`anomaly` and `climatology` are both deprecated in this PR for future removal.

### DIFF
--- a/docs/user_api/index.rst
+++ b/docs/user_api/index.rst
@@ -12,9 +12,8 @@ Climatologies
    :nosignatures:
    :toctree: ./generated/
 
-   anomaly
    calendar_average
-   climatology
+   climate_anomaly
    climatology_average
    month_to_season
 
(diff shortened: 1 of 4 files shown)


hCraker:
The link check is going to fail because I link to the new function in the `anomaly` docstring.

jukent (approved):
Looks great :)

PR #335, opened 2023-01-31
Allow upstream-dev to run on forks

anissa111:
Summary of changes:
- remove condition of only allowed to be run on NCAR/geocat-comp repo


hCraker (approved):
Thanks for doing this!

PR #336, opened 2023-01-31
Upstream CI

anissa111:
Summary of changes:
- re-arrange order of action to set up conda, install pip dependencies, then install conda dependencies
- remove pip installs from conda env file
- switch to using miniconda instead of mamba
- rename the upstream dev environment to be distinct from the development environment
- update to run on python 3.10 instead of 3.8

See passing upstream CI here: https://github.com/anissa111/geocat-comp/actions/runs/4049852058


hCraker (approved):
Looks good once the conflicts with main are resolved! Thanks

erogluorhan (approved):

PR #337, opened 2023-01-31
Add release notes to the webpage and add formatting for auto generated notes

hCraker:


hCraker:
Whoops! I don't know how forks work apparently 😅

docs/index.rst
```diff
@@ -82,7 +82,14 @@ novel computational functions for geosciences data.
    API Reference <api>
    Usage Examples <examples>
    Cite GeoCAT-comp <citation.rst>
+
+.. toctree::
+   :maxdepth: 2
+   :hidden:
+   :caption: Extra Information
```
anissa111, line 89: Could we call this section something like "For Developers"? I think we'd discussed something similar in the past if I remember correctly

We could also link relevant contributors guides and the published roadmap from here, too.

Also, I'm not sure that this is where the NCL link should go. I think non-developers are more likely to be interested in (and need a link to) the NCL docs

docs/release-notes.rst
```diff
@@ -0,0 +1,52 @@
+.. currentmodule:: geocat.comp
+
+.. _release:
+
+v2023.02.0
+==========
```
anissa111, line 6: This structure is making the in-doc toctree act a little funny.

image

It's titling the whole page with the release number and populating the right side toc tree with the subheaders.

I suggest making the top level header something like "Release Notes" and decreasing the heading level for existing headers

docs/conf.py
```diff
@@ -77,6 +78,11 @@ def __getattr__(cls, name):
     'pint': ('https://pint.readthedocs.io/en/stable/', None),
     'cftime': ('https://unidata.github.io/cftime/', None),
 }
+# allows us to easily link PRs and issues in the change log
+extlinks = {
+    "issue": ("https://github.com/NCAR/geocat-comp/issues/%s", "GH#"),
+    "pr": ("https://github.com/NCAR/geocat-comp/pull/%s", "GH#"),
```
anissa111, line 84:
```suggestion
    "issue": ("https://github.com/NCAR/geocat-comp/issues/%s", "GH"),
    "pr": ("https://github.com/NCAR/geocat-comp/pull/%s", "PR"),
```

Could we do this to distinguish between issues and pull requests? (This is what xarray does)

docs/release-notes.rst
```diff
@@ -0,0 +1,52 @@
+.. currentmodule:: geocat.comp
+
+.. _release:
```
anissa111, line 3: How automatic is this page btw? Is this something that we'll need to update with every release (I'm okay with that, just making sure I understand)?

docs/release-notes.rst
```diff
@@ -0,0 +1,52 @@
...
+
+v2023.02.0
```
anissa111, line 5: It might be reasonable to recreate the release notes from the last few releases here, too

docs/release-notes.rst
```diff
@@ -0,0 +1,52 @@
...
+^^^^^^^^^^^
+* Adding import nightly CI test by `Anissa Zacharias`_ in (:pr:`300`)
+* Expand upstream CI by `Anissa Zacharias`_ in (:pr:`301`)
+* Patch for import test by `Anissa Zacharias`_ in (:pr:`302`)
+* Fix the import package CI test by `Anissa Zacharias`_ in (:pr:`303`)
+* CI failures quick patch by `Anissa Zacharias`_ in (:pr:`312`)
+* Resolve CI link-check issue by `Anissa Zacharias`_ in (:pr:`316`)
+* Pin numpy for numba compatibility by `Anissa Zacharias`_ in (:pr:`325`)
+* Bump actions/setup-python from 4.3.0 to 4.4.0 by `Dependabot`_ in (:pr:`322`)
+* Bump pypa/gh-action-pypi-publish from 1.5.1 to 1.6.4 by `Dependabot`_ in (:pr:`317`)
```
anissa111, line 38: Not sure we need to include dependabot updates, especially if they're not major

docs/release-notes.rst
```diff
@@ -0,0 +1,52 @@
...
+* Fastertest by @pilotchute in (:pr:`327`)
```
anissa111, line 39: What do you think about adding a little explanation to bullet points that are either really influential or don't have a super obvious title?

anissa111 (changes requested):

docs/release-notes.rst
anissa111, line 5:
```suggestion
v2023.02.0 (Feb 2, 2023)
```

We should also probably include the date

docs/release-notes.rst
hCraker: Yeah I wasn't sure about this either. I'm gonna remove them

docs/release-notes.rst
hCraker: Yeah we could do that. How far back do you think we should go?

anissa111: I'm not sure. Maybe [v2022.10.0](https://github.com/NCAR/geocat-comp/releases/tag/v2022.10.0)?

docs/index.rst
hCraker: Yeah we should definitely make this a "For Devs" section. The release note, contributors guide, and the roadmap would fit in here nicely. I'll put the NCL link in the users section

docs/conf.py
hCraker: Sure! I was just modeling this off of what Dask does for their release notes formatting. But a distinction is a good idea. Although, I'm not sure when an issue will ever make it into the release notes, but one could be added elsewhere

docs/release-notes.rst
hCraker: I added `.github/release.yml` to format the autogenerated release notes on GitHub. Then we can copy those to this `.rst` file and modify the formatting for usernames and PR links by using find-and-replace. It will require only a little bit of manual work, but it should be fairly quick.

docs/conf.py
anissa111: Totally fair. I think it might be relevant to link to an issue if significant discussion happened on it, but yeah you're right it might not actually be used.

docs/release-notes.rst
anissa111: Okay, neat. That's good enough for me. It will also allow us to add any extra context and links on both the improved release notes here on github tags and the release notes docs

docs/release-notes.rst
hCraker: I can for sure do that since I work on geocat-comp. The only concern is I'm not sure I could hold that same standard when I make release notes for other packages or for comp PRs that I wasn't really involved with. I can add details to the comp release notes and ask for help on release notes for other packages if we decided to add a "Release Notes" page to the other docs.

docs/conf.py
hCraker: It certainly doesn't hurt to add it, so I'll go ahead and commit this

docs/release-notes.rst
hCraker: That seems doable. It's only two releases back

PR #338, opened 2023-02-01
Adding method to calculate delta pressure

jukent:
Includes tests
I need to add wrappers so that the original NCL function name (dpres_plev) also works. I could use some assistance with that step.


jukent:
Sorry for so many commits. I'm having some weird import errors when running the tests locally that I haven't figured out yet

anissa111:
You're good! We can unsubscribe if we need to. Do you want any help with whatever's happening on your local?

jukent:
The link check failure doesn't seem to be connected to this PR

hCraker:
> The link check failure doesn't seem to be connected to this PR

You're right! It's due to something I added. It will be resolved once the release is made today and the updated documentation is live.

src/geocat/comp/meteorology.py
```diff
@@ -1431,3 +1431,155 @@ def saturation_vapor_pressure_slope(
...
+        pressure_lev = np.flip(pressure_lev)
+
+    # Calculate delta pressure
+    delta_pressure = np.empty_like(pressure_lev)
+
+    delta_pressure[0] = (pressure_lev[0] +
+                         pressure_lev[1]) / 2 - pressure_top  # top level
+    for i in (np.arange(1, len(pressure_lev) - 1)):  # middle levels
+        delta_pressure[i] = (pressure_lev[i + 1] - pressure_lev[i - 1]) / 2
+        i += 1
```
anissa111, line 1482: I think this will also need some vectorization for future dask integration

anissa111 (commented):
Okay, I've left a few small comments down below and I have a few broader comments here:
1. I added your functions to the internal and user API when I made the NCL function name wrapper section. Feel free to add those functions when available there.
1. What do you think about calling these functions `delta_pressure` and `delta_pressure1D` or similar? I'm not sure the `calc_` is necessary, as we're not doing it elsewhere in -comp for similar functions (@hCraker thoughts?)

hCraker:
> 2. What do you think about calling these functions `delta_pressure` and `delta_pressure1D` or similar? I'm not sure the `calc_` is necessary, as we're not doing it elsewhere in -comp for similar functions (@hCraker thoughts?)

I agree with this. I think `delta_pressure` is sufficiently descriptive.

jukent:
Thank you for your review. I'll address these in a bit.

jukent:
Thank you for your review. I'll address these in a bit.

src/geocat/comp/meteorology.py
```diff
@@ -1431,3 +1431,155 @@ def saturation_vapor_pressure_slope(
             tfill)
 
     return svp_slope
+
+
+def _calc_deltapressure_1D(pressure_lev, surface_pressure):
+    """Helper function for `calc_deltapressure1. Calculates the pressure layer
```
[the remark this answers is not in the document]
jukent: huh it let me update your comment to commit the change as I want it. I didn't expect that.

src/geocat/comp/meteorology.py
```diff
@@ -1431,3 +1431,155 @@ def saturation_vapor_pressure_slope(
...
+
+def calc_deltapressure(pressure_lev, surface_pressure):
+    """Calculates the pressure layer thickness (delta pressure) of a constant
+    pressure level coordinate system.
+
+    Returns an array of length matching `pressure_lev`.
+
+    Parameters
+    ----------
+    pressure_lev : :class:`np.Array`, :class:'xr.DataArray`
```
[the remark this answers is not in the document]
jukent: I didn't even realize they were supposed to be autolinks. I just mirrored it from elsewhere. I'll change this.

src/geocat/comp/meteorology.py
```diff
@@ -1431,3 +1431,155 @@ def saturation_vapor_pressure_slope(
...
+        dims = len(surface_pressure.shape)
+    except:
+        dims = 0
+
+    # Safety check
+    if dims > 3:
+        warnings.warn("`surface_pressure` cannot have more than 3 dimensions.")
+
+    # Convert to floats to prevent integer division rounding errors
+    pressure_lev = [float(i) for i in pressure_lev]
```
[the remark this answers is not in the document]
jukent: So the reason I did it here, is because the same pressure_lev array is used for every single surface_pressure array element - so we would be converting it multiple times if we did it at the point where we use it. As opposed to just once.  I can say with some confidence that the user should not be chunking along the pressure_lev, but along some dimension of their surface_pressure.

If there is a different way to enforce that the division is float division, that could be an option?

src/geocat/comp/meteorology.py
jukent: Similarly, I don't expect a user to be chunking along the one-dimensional pressure_lev, but along a dimension of surface_pressure. But I guess they always could find a reason? I don't know much about vectorization for dask integration tbh. I could use some help with that.

jukent:
Okay I'll change the name to `delta_pressure` and `delta_pressure1D` throughout. I just made it start with `calc_` because of the opinion that all methods should be verbs for readability. It does make everything longer though. Working on this now.

src/geocat/comp/meteorology.py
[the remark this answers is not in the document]
jukent: I'm going to remove the calls to float in my next PR and hope that it treats `/` as float division as it is supposed to by default. The test will pass if it doesn't, and maybe that is worth investigating more than this potentially slow mapping is.

jukent:
To do:
[] Dask integration vectorization -- help please
[] NCL wrapper methods (and add to NCL section of docs)
[] Better duck typing (I've never done this before either)

src/geocat/comp/meteorology.py
```diff
@@ -1431,3 +1431,155 @@ def saturation_vapor_pressure_slope(
...
+    # Convert inputs to numpy arrays
+    try:
+        pressure_lev = np.asarray(pressure_lev)
+    except AttributeError:
+        print("`pressure_lev` must be array-like.")
+
+    try:
+        surface_pressure = np.asarray(surface_pressure)
+    except AttributeError:
+        print("`surface_pressure` must be array-like.")
```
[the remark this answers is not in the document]
jukent: I haven't done duck typing before so I could use some help on this. Maybe during team time. I just removed the call convert to numpy and had to change `pressure_lev` to `pressure_lev.values` when adding it as a coordinate. It is still passing tests so yay

jukent:
dpres_plev_preview
dpres_plev_doc

Huh the docs aren't rendering how I'd expect. I'll iterate on this.

anissa111:
Ah! Yeah that's an .rst quirk. See https://jwodder.github.io/kbits/posts/rst-hyperlinks/
image

anissa111:
We actually do this to reduce the amount of warnings that get printed when the docs generate
image

jukent:
Thanks that helps with the links. I'm still not sure why there is the funny indentation on the dpres_plev page. Probably related to how I'm combining two docstrings, but it isn't where the join is.

jukent:
Screen Shot 2023-02-03 at 12 59 22 PM
Links fixed, still this weird indenting.

hCraker:
So I'm not sure if this is the cleanest way of doing things, but I did fix the formatting issue @jukent was having. Sphinx is *very* picky about white space, and the string formatting prefix (`f"..."`) saves all white space as is. This means the new line characters added to make the code more readable ended up in the documentation and causing the indentation.

I did try a few things to see if I could copy the doc string from `delta_pressure` without having to manually set the `__doc__` attribute for `dpres_level` with no luck. I can explore some more if you'd like.

jukent:
Thanks @hCraker ! I tried a few different things, and pre-commit would always undo it. Thanks so much for finding the right combination. I still don't fully understand why the formatting break was in the middle of the original docstring (but at a line break) - but I'm glad this works.

jukent:
So the NCL fx this is based on specifies that it is for a constant pressure level system. As in every coordinate we have the same pressures as dimensions. The file I am working with for the larger VIMFC workflow has different pressure system over every lat-lon-time coordinate. Should we change this to work for that kind of system?

src/geocat/comp/meteorology.py
anissa111: You know what, that's true enough though. Let's just make sure nothing in here will force a compute. 

And we can probably re-arrange this to be a little more pythonic, though. Something like 

```suggestion
    delta_pressure[1:-1] = [(a-b)/2 for a, b in zip(pressure_lev[2:], pressure_lev[:-1])]
```

anissa111:
> So the NCL fx this is based on specifies that it is for a constant pressure level system. As in every coordinate we have the same pressures as dimensions. The file I am working with for the larger VIMFC workflow has different pressure system over every lat-lon-time coordinate. Should we change this to work for that kind of system?

For documenting this on the PR in addition to our conversation yesterday: no, I don't think we need to have this now. This functionality would be cool to have later down the line, but for now we should be able to release the base functionality without this extra.

anissa111:
Also, we're actually using the "developer feature" for development standardization and like best practices updates. This would fall under the "feature" tag

src/geocat/comp/meteorology.py
```diff
@@ -1431,3 +1431,154 @@ def saturation_vapor_pressure_slope(
...
+                                      name="delta pressure")
+
+    return delta_pressure
+
+
+def dpres_plev(pressure_lev, surface_pressure):
+    return delta_pressure(pressure_lev, surface_pressure)
+
+
+_dpres_plev_doc_str = f"""This method is a wrapper for `delta_pressure <https://geocat-comp.readthedocs.io/en/stable/user_api/generated/geocat.comp.meteorology.delta_pressure.html>`_. {delta_pressure.__doc__}"""
```
anissa111, line 1583: Could we (or, should we?) add a new line between the wrapper statement and the original docstring? 
image

hCraker: I think you can with a `/n` newline character. I'm not what sphinx will do with that, but we can try it

anissa111: Yeah, I think that would work.

jukent: Hmm I added `\n` and it is ignored. Do you think this is an f-string issue or a readthedocs issue?

jukent:
Link checker is failing with a link that will only exist after this PR is merged: https://geocat-comp.readthedocs.io/en/stable/user_api/generated/geocat.comp.meteorology.delta_pressure.html

jukent:
I believe the next step is to remove the geocat-f2py version of dpres_plev and just leave a warning? @anissa111 you said you would do this - do you want to pair program it next week so I can be equipped for the next PR that deprecates something?

anissa111:
> Link checker is failing with a link that will only exist after this PR is merged: https://geocat-comp.readthedocs.io/en/stable/user_api/generated/geocat.comp.meteorology.delta_pressure.html

@jukent and @hCraker, there's a better way to be doing this. Wherever possible we should try to have sphinx generate the links itself upon build, which will avoid this kind of issue and also make sure that previous versions link internally to themselves instead of the latest version. 

Don't worry about it here because I'm making a helper function that'll take care of it, but ideally we shouldn't ever have to rely on links that don't exist yet for internal sphinx linking.

jukent:
So in general, internal links are preferred yeah? I won't worry about it this time.

anissa111:
> So in general, internal links are preferred yeah? I won't worry about it this time.

Yup! For something like this, it'd be like using the :func: markup. For other documentation pages, the :ref: markup. Link-links should probably only be used to refer to things outside of our own documentation.

src/geocat/comp/meteorology.py
```diff
@@ -1431,3 +1431,154 @@ def saturation_vapor_pressure_slope(
...
+    Returns an array of shape matching (`surface_pressure`, `pressure_lev`).
+
+    Parameters
+    ----------
+    pressure_lev : :class:`numpy.ndarray`, :class:'xarray.DataArray`
+        The pressure level array. May be in ascending or descending order.
+        Must have the same units as `surface_pressure`.
+    surface_pressure : :class:`np.Array`, :class:'xr.DataArray`
+        The scalar or N-dimensional surface pressure array. Must have the same
+        units as `pressure_lev`. Cannot exceed 3 dimensions.
```
hCraker, line 1506: Given how xarray operates, is this truly limited to 3D or is that from text you copied over from the Fortran/f2py wrapper docs?

hCraker (commented):
I haven't yet looked through the test file, but I have some initial comments to post before a meeting

src/geocat/comp/meteorology.py
```diff
@@ -1431,3 +1431,154 @@ def saturation_vapor_pressure_slope(
...
+    -------
+    delta_pressure : :class:`numpy.ndarray`
+        The pressure layer thickness array. Shares dimensions and units of
+        `pressure_lev`.
+    """
+    pressure_top = min(pressure_lev)
+
+    # Safety checks
+    if not surface_pressure:
+        warnings.warn("'surface_pressure` can't equal a missing value.")
```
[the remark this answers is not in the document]
jukent: True.

src/geocat/comp/meteorology.py
jukent: It is not truly limited - just physically (lat lon and time). But perhaps we want to let the user find some novel 5-D surface dimensions to apply it to? I think this was hard coded in the NCL version because of the for-loops. Should I take this out?

jukent: If I take this out, there is no need to know the dimensions, just whether it is scalar or not, so I think this part of the code could be reworked

```
    # Get dimensions of `surface_pressure`
    try:
        dims = len(surface_pressure.shape)
    except:
        dims = 0

    # Calculate delta pressure
    if dims == 0:  # scalar case
        delta_pressure = _delta_pressure1D(pressure_lev, surface_pressure)
    else:  # multi-dimensional cases
        . . .
```

jukent: using np.isscalar to check.

jukent:
Failures seem to be due to some tests not being deleted that were on now deleted (formally deprecated) methods such as 'geocat.comp.skewt_params'. I can look into this if you want in a separate PR?

hCraker:
> Failures seem to be due to some tests not being deleted that were on now deleted (formally deprecated) methods such as 'geocat.comp.skewt_params'. I can look into this if you want in a separate PR?

This is fixed on the main branch, so merging the changes from main should fix this. I'm not sure why GitHub isn't prompting us to merge main into this PR though.

test/test_meteorology.py
```diff
@@ -16,14 +16,14 @@
...
+        delta_pressure)
 else:
-    from geocat.comp.meteorology import (dewtemp, heat_index, relhum,
-                                         relhum_ice, relhum_water,
-                                         actual_saturation_vapor_pressure,
-                                         max_daylight, psychrometric_constant,
-                                         saturation_vapor_pressure,
-                                         saturation_vapor_pressure_slope)
+    from geocat.comp.meteorology import (
+        dewtemp, heat_index, relhum, relhum_ice, relhum_water, showalter_index,
```
hCraker, line 23:
```suggestion
        dewtemp, heat_index, relhum, relhum_ice, relhum_water
```

hCraker (changes requested):
This should fix the failures

src/geocat/comp/__init__.py
```diff
@@ -8,7 +8,8 @@
 from .meteorology import (dewtemp, heat_index, relhum, relhum_ice, relhum_water,
                           actual_saturation_vapor_pressure, max_daylight,
                           psychrometric_constant, saturation_vapor_pressure,
-                          saturation_vapor_pressure_slope)
+                          saturation_vapor_pressure_slope, delta_pressure)
+from .skewt_params import get_skewt_vars, showalter_index
```
hCraker, line 12: Delete trying to import something that doesn't exist
```suggestion
```

hCraker (changes requested):

test/test_meteorology.py
```diff
@@ -16,14 +16,14 @@
...
+        delta_pressure)
 else:
-    from geocat.comp.meteorology import (dewtemp, heat_index, relhum,
-                                         relhum_ice, relhum_water,
-                                         actual_saturation_vapor_pressure,
-                                         max_daylight, psychrometric_constant,
-                                         saturation_vapor_pressure,
-                                         saturation_vapor_pressure_slope)
+    from geocat.comp.meteorology import (
+        dewtemp, heat_index, relhum, relhum_ice, relhum_water
```
hCraker, line 23:
```suggestion
        dewtemp, heat_index, relhum, relhum_ice, relhum_water,
```

jukent:
For anyone else - the problem was a git conflict that I incorrectly resolved

hCraker (approved):
Looks good!

PR #339, opened 2023-02-02
adjust calendar example to follow similar template

jukent:


jukent:
Perhaps I should look into setting up a reviewNB for the examples PRs.

anissa111:
> Perhaps I should look into setting up a reviewNB for the examples PRs.

Yeah, that would be super useful! Currently we're just looking at the generated documentation, but something like reviewNB could really help us review new example _content_, too!

jukent:
I just sent a request for Review NB to have access to this repository. Keep an eye out for an email I guess?

docs/examples/calendar_average.ipynb
```diff
@@ -9,41 +9,41 @@
...
         "## Import packages\n",
-        "\n"
+        "\n",
+        "Dependencies:\n",
+        "\n",
+        "- geocat.comp\n",
+        "- geocat.datafiles (for accessing data file only)\n",
+        "- cftime (installed with geocat.comp)\n",
+        "- matplotlib (installed with geocat.viz)\n",
+        "- xarray (installed with geocat.comp)"
```
hCraker, line 28: I think it'd be good practice to link to our dependency's documentation pages.
```suggestion
        "Dependencies:\n",
        "\n",
        "- geocat.comp\n",
        "- [geocat.datafiles](https://github.com/NCAR/geocat-datafiles) (for accessing data file only)\n",
        "- [cftime](https://unidata.github.io/cftime/) (installed with geocat.comp)\n",
        "- [matplotlib](https://matplotlib.org/) (installed with geocat.viz)\n",
        "- [xarray](https://docs.xarray.dev/en/stable/) (installed with geocat.comp)"
```

hCraker (changes requested):
Just a small change. Otherwise, I think this looks good!

docs/examples/calendar_average.ipynb
```diff
@@ -9,41 +9,41 @@
...
       "metadata": {
         "collapsed": false
       },
-      "outputs": [],
+      "outputs": [
+        {
+          "name": "stderr",
+          "output_type": "stream",
+          "text": [
+            "/Users/jkentadmin/miniconda3/envs/geocatsandbox/lib/python3.10/site-packages/geocat/viz/__init__.py:7: UserWarning: GeoCAT packages will no longer be published on the ncar channel, switch to conda-forge\n",
```
hCraker, line 42: Wait we should be running this in such a way where we don't get this warning at all. This seems to be a viz specific problem, but we should try to figure out how to avoid this warning and demonstrate best practices.

jukent: huh, if I clear all output and push -- will it go away? I didn't expect that to appear on the GH build tbh

jukent: I don't want to put a warning suppression in the examples - that seems not great. Will it go away if I install with conda-forge instead? I'll note that the install instructions for geocat-viz still say to use the ncar channel.

jukent: Removed viz, should be good on this warning now

hCraker (changes requested):
Thanks for removing viz. Once difference I noticed is that in the [current implementation](https://geocat-comp.readthedocs.io/en/stable/examples/calendar_average.html#plot) of the plot that is live online has years for the x-axis where your version has something that is less human readable. `nc-time-axis` is what I've used in the past to format datetimes for labels. See [this example](https://github.com/NCAR/geocat-tutorials/blob/a3a02211b946ff105d7ccdd1dd07667d6195736d/notebooks/geocat-comp/02-more_advanced_functionality.ipynb) from the geocat-comp tutorial last November.

You should just need to import `nc_time_axis` for it to work.

hCraker:
I just checked and `nc-time-axis` is not currently included in our dependencies for the documentation. You can add it to `build_envs/docs.yml` to get this to work

jukent:
Okay thanks @hCraker. I wasn't too concerned with making the plot pretty, so I didn't event think about that. The example you shared imports `nc_time_axis` but doesn't seem to do anything with it. It should be done now

hCraker:
> The example you shared imports `nc_time_axis` but doesn't seem to do anything with it. It should be done now

Yeah you don't need to call it once it's imported. It integrates with matplotlib in the background from my understanding. Thanks for adding it!

jukent:
@hCraker I tried that and it didn't change the plot at all, so I just added to lines to call it.

hCraker:
Huh I guess I'm mistaken. I'll look into this next week. I honestly can't remember my process for the tutorial script.

jukent:
@anissa111 and @hCraker I think this is ready for review for real and incorporates the feedback received in the climatology PR.

hCraker (approved):
I found a typo, but otherwise this looks great! 🎉

jukent:
I think it's ready

PR #341, opened 2023-02-02
add climatology_average example

jukent:


hCraker (commented):
Don't forget to add the example to `docs/examples.rst` so that it shows up on the website. Once you do that, I can review this.

hCraker:
I forgot to mention that you also have to add your example to `docs/notebook-examples.txt` for the card to show up. Sorry!

hCraker:
To make the card for the new example, you'll need a thumbnail. I'd recommend using a placeholder for now or using the thumbnail from the other example. Once this PR is all done and the plot is in its final state, then you can make a more appropriate thumbnail.

jukent:
I'm not seeing that file anywhere in the repository?

jukent:
Oh it's in the `.gitignore`? Should we change that? That is unexpected unless that file is rebuilt by GH.

anissa111:
@jukent @hCraker It's not actually `docs/notebook-examples.txt` you need to add to. That's a file that's generated/remade when the docs are made. 

You'll actually want to add it as an entry in `docs/gallery.yml` to add it to the gallery page with an appropriate thumbnail.

jukent:
That makes sense. Thanks @anissa111 and @hCraker. I will do that.

hCraker:
Whoops! Thanks for the clarification @anissa111. I was doing me best to infer based on your last examples PR, but I missed a few things

anissa111:
Another thing to add to the PR standards checklist!

jukent:
> 

@hCraker stop replying to GitHub you're supposed to have the day off 💻

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,193 @@
...
+        "ax[0].plot(time_num_raw, temp.data)\n",
+        "ax[0].set_ylabel('Raw Data')\n",
+        "\n",
+        "ax[1].plot(time_num_day, daily.data)\n",
+        "ax[1].set_ylabel('Daily Climatology')\n",
+        "\n",
+        "ax[2].plot(time_num_month, monthly.data)\n",
+        "ax[2].set_ylabel('Monthly Climatology')\n",
+        "\n",
+        "ax.set_xlabel('Time')\n",
```
hCraker, line 157: I'm getting an attribute error here since `ax` is a list of axes objects. 
`AttributeError: 'numpy.ndarray' object has no attribute 'set_xlabel'`

Changing this line to add the label to the bottom set of axes only is how you can get the label to show up properly.

```suggestion
        "ax[2].set_xlabel('Time')\n",
```

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,193 @@
...
+        "\n",
+        "# Add title manually to control spacing\n",
+        "title_str = 'Climatology Average on 6-hourly Data \\n ' + temp.long_name + ' (' + temp.units + ')'\n",
+        "fig.suptitle(title_str, fontsize = 20)\n",
+        "\n",
+        "plt.show();"
```
hCraker, line 163: This figure doesn't show up on the rendered notebook. Did you render this block before pushing your changes? Also, the `;` shouldn't be needed
```suggestion
        "plt.show()"
```

hCraker (changes requested):
Pretty good just some formatting changes and a bug.

docs/gallery.yml
```diff
@@ -1,3 +1,7 @@
 - title: Calendar Averaging
   path: examples/calendar_average.ipynb
   thumbnail: _static/thumbnails/sphx_glr_calendar_average_example.png
+
+- title: Climatology Average
+  path: examples/climatology_average.ipynb
+  thumbnail: _static/thumbnails/sphx_glr_calendar_average_example.png
```
anissa111, line 7: This example should have its own thumbnail

jukent: Yes, sorry I just put a placeholder there while I make a new one. All good points. 

I think the value of listing dependencies in markdown is linking documentation. We could move this to the end of the script, like a read more section after the example?

jukent:
Hi could you take another look at this now that I've incorporated @anissa111 's feedback.

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,208 @@
...
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "## Read in data\n",
+        "\n",
+        "We will get the data from the [geocat-datafiles](https://github.com/NCAR/geocat-datafiles) package. This package contains example data used in many of the examples for geocat packages.\n",
+        "\n",
+        "Then, we use xarray's [open_dataset](http://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html) function to read the data into an xarray dataset, choose a single model from the ensemble run, and extract its surface temperature data into `temp`"
```
hCraker, line 38:
```suggestion
        "Then, we use xarray's [open_dataset](http://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html) function to read the data into an xarray dataset, choose a single model from the ensemble run, and extract its surface temperature data into `temp`."
```
Missing punctuation

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,208 @@
...
+    {
+      "attachments": {},
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "## Look at the Raw Data\n",
+        "\n",
+        "Before we compute the climatologies, let's take a look at the raw surface temperature data from our selected model run with temporal resolution of 6-hours.\n",
+        "\n",
+        "The plot output has adjusted datetimes instead of using integers to denote the day of the year for the time axis. The year for the outputted data is the floor of the median year of the inputted data, which is 1993 in this case. This will be repeated for all subsequent plots in this example."
```
hCraker, line 64: > The plot output has adjusted datetimes instead of using integers to denote the day of the year for the time axis.

I'm a bit confused about this part since the plot has integers for the time axis. It looks like you're doing the conversion, but is isn't showing up properly in the figure 

Edit: It looks like this is an issue with a few different figures

image

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,208 @@
...
+              "<Figure size 640x480 with 1 Axes>"
+            ]
+          },
+          "metadata": {},
+          "output_type": "display_data"
+        }
+      ],
+      "source": [
+        "# Compute\n",
+        "daily = climatology_average(temp, 'day')\n",
```
hCraker, line 133: You mention in the text above this cell that you are setting the `freq` argument to "daily". While we can use positional arguments here, can you change this for clarity?

```suggestion
        "daily = climatology_average(temp, freq='day')\n",
```

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,208 @@
...
+        "\n",
+        "Note that while daily and monthly climatology averages are demonstrated here, you could also use the frequency keyword argments `hour` or `season` to calculate hourly or seasonal climatology averages."
+      ]
+    },
+    {
+      "attachments": {},
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "Let's start with the daily climatology average. The following plot shows output of the raw data being aggregated using the `climatology_average` function with the `freq` argument set to 'daily'. "
```
hCraker, line 112: Typo

```suggestion
        "Let's start with the daily climatology average. The following plot shows output of the raw data being aggregated using the `climatology_average` function with the `freq` argument set to 'day'. "
```

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,208 @@
...
+        "plt.title('Daily Climatology')\n",
+        "plt.xlabel(\"Time\");"
+      ]
+    },
+    {
+      "attachments": {},
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "Next we'll compute the monthly climatology averages by using `climatology_average` with the `freq` argument set to `monthly`.\n",
```
hCraker, line 148: Similar typo as above. The argument value is "month" rather than "monthly". This brings up an interesting thought though. Should `freq` take in the values "hourly", "daily", "monthly", and "seasonal" rather than the current "hour", "day", "month", "season" since it would be more intuitive and closer to natural speech? @anissa111 @jukent Thoughts?
```suggestion
        "Next we'll compute the monthly climatology averages by using `climatology_average` with the `freq` argument set to `month`.\n",
```

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,208 @@
...
...nw8vISzs7O4p577hFpaWkWbXJzc8Wjjz4q1Gq1UKvV4tFHHxV5eXkWba5cuSKmT58unJ2dhZeXl1iyZIkoKytrjbfe7jXm3mRmZoro6GgRGBgonJycRHh4uHjnnXeE0WiU2vDetLwFCxaIkJAQoVQqha+vr5g0aZLFF7fRaBSvvvqq0Gq1QqVSibvuukskJCRYnIP3pfXUd39SUlLq/Le1d+9e6Ry8P62joX87t6st8eG9aX0yIYRo614mIiIiIlvo0DU+RERERLdi4kNERESdBhMfIiIi6jSY+BAREVGnwcSHiIiIOg0mPkRERNRpMPEhIiKiToOJDxF1GKtWrcKgQYNsHQYRtWNcwJCI7EJDu0rPnz8fa9euhcFgqLE/FRGRGRMfIrILt+42vWXLFvz5z3+22GDY2dkZGo3GFqERkR3hUBcR2QWtVis9NBoNZDJZjeduH+qKjo7G7Nmz8cYbb8Df3x8eHh547bXXUFlZiZdeegleXl4ICgrCv//9b4trXb16FXPnzoWnpye8vb0xa9YspKamtu0bJqJWwcSHiDq0mJgYXLt2Dfv27cO7776LVatW4Z577oGnpyeOHDmChQsXYuHChUhPTwcAlJSUYMKECXBzc8O+fftw4MABuLm5Ydq0aSgvL7fxuyGi5mLiQ0QdmpeXF/7+978jPDwcCxYsQHh4OEpKSrBixQr06tULy5cvh1KpxMGDBwEAmzdvhlwux6effor+/fujT58+WL9+PdLS0hAbG2vbN0NEzeZg6wCIiFpTv379IJff/BvP398fERER0s8KhQLe3t7Izs4GABw/fhyXLl2CWq22OE9ZWRmSk5PbJmgiajVMfIioQ3N0dLT4WSaT1fqc0WgEABiNRgwdOhRffvlljXP5+vq2XqBE1CaY+BAR3WLIkCHYsmUL/Pz84O7ubutwiKiFscaHiOgWjz76KHx8fDBr1izs378fKSkpiIuLw7PPPouMjAxbh0dEzcTEh4joFi4uLti3bx+6du2Ke++9F3369MGCBQtQWlrKHiCiDoALGBIREVGnwR4fIiIi6jSY+BAREVGnwcSHiIiIOg0mPkRERNRpMPEhIiKiToOJDxEREXUaTHyIiIio02DiQ0RERJ0GEx8iIiLqNJj4EBERUafBxIeIiIg6DSY+RERE1Gn8f4mL0Waewc9XAAAAAElFTkSuQmCC",
```
hCraker, line 160: The images for daily and monthly climatologies look the same other than the title. Is this an error?

Daily:
image

Monthly:
image

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,208 @@
...
+              "<Figure size 640x480 with 1 Axes>"
+            ]
+          },
+          "metadata": {},
+          "output_type": "display_data"
+        }
+      ],
+      "source": [
+        "# Compute\n",
+        "monthly = climatology_average(temp, 'month')\n",
```
hCraker, line 171: Similar suggestion as seen above. I think we should use the name of the argument rather than using the positional option for clarity in example scripts. I can see an argument for going either way, but I think this will avoid any confusion.
```suggestion
        "monthly = climatology_average(temp, freq='month')\n",
```

docs/examples/climatology_average.ipynb
anissa111: I'm not sure I have a strong opinion about this. I think I would err on the side of following what similar packages are using. Is there any precedent in scipy, astropy, xarray, or anything similar?

hCraker: I don't think there is a precedent since I designed this functions to fill a need that wasn't being met by other packages. It wouldn't be too hard to add `monthly` as an alias for `month` with how the function already handles inputs. It might be a nice quality of life thing, but is low priority

jukent: I think it's fine as is though. I just didn't notice that I had done it incorrectly because no errors or warnings were raised. Perhaps something should be raised if the frequency is set to a value the method doesn't know?

anissa111 (commented):
This is looking really good! Thanks for incorporating my suggestions about moving away from an -examples style script. I've left a few small points below along with Heather's suggestions.

Additionally, this is a small point, but we could still include all three plots in the thumbnail like before even though we technically never plot them all together.

docs/examples/climatology_average.ipynb
jukent: I found the problem. I'll push that and the subplot gallery card.

jukent: well the problem with the subplots but not with the x axis.

docs/examples/climatology_average.ipynb
hCraker: > I think it's fine as is though. I just didn't notice that I had done it incorrectly because no errors or warnings were raised. Perhaps something should be raised if the frequency is set to a value the method doesn't know?

You had the value set correctly in the function call but had a typo in the explanatory text. That kind of argument check does exists, but luckily you did things a-ok and didn't trigger it.

docs/examples/climatology_average.ipynb
```diff
@@ -0,0 +1,208 @@
...
+        "collapsed": false
+      },
+      "outputs": [],
+      "source": [
+        "import cftime\n",
+        "import matplotlib.pyplot as plt\n",
+        "import xarray as xr\n",
+        "\n",
+        "from geocat.comp import climatology_average\n",
+        "import geocat.datafiles as gdf"
```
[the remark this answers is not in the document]
hCraker: Ooo yes I like that. We have done that before in various tutorials and should probably do it here too

docs/examples/climatology_average.ipynb
jukent: fixed the x-axis too. It was some script I dropped when removing gc-viz.

docs/examples/climatology_average.ipynb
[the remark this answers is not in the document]
jukent: Done

jukent:
Unrelated broken link:

`(    installation: line   49) broken    https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html - 404 Client Error: Not Found for url: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html`

hCraker (approved):
Thanks for adding this example!

anissa111 (approved):

PR #343, opened 2023-02-03
first pass at fourier filter example

jukent:
First pass at changing the [fourier filter example](https://geocat-examples.readthedocs.io/en/latest/gallery-geocat-comp/matplotlib/fourier_filter_example.html#sphx-glr-gallery-geocat-comp-matplotlib-fourier-filter-example-py) into a notebook. 

I have a lot of questions for @pilotchute that I left in the document. The answers to these questions should be self evident to the reader:
- what is the difference between your three plots (none of them have titles)
- what is the 1 and 2 tide frequency cut offs that you chose to do analysis on? Are they arbitrary?
- what is the variable of the data?


jukent:
I don't see the reason to demonstrate 2 different fourier filters - as in I'm not sure what is unique about the different cases to make it interesting. So I am simplifying the example to just one filter.

To be honest, the original example is confusing to me: Just the nature of having multiple subplots within one plot (so the counting restarts at 0 in the middle of the code block) make it hard to track which code maps to which plot, variables are overwritten, the plots aren't well labeled/don't have legends on multi-line plots, and there isn't much explanatory text. If I miss something that you think is missing from the original please add it -- but if the example is just to demonstrate the use of the fourier filter function, this trimmed down version accomplishes that.

Also the dataset used has dimensions dim0 and dim1, and since I don't know much about this dataset I don't know what a better name would be. So I didn't both adding labels to the plots yet because I don't know what to label them. I can look into this, but @pilotchute since you wrote the example originally, could you answer this question quicker?

I have a lot of questions: What is this dataset of? It doesn't seem like geoscience data (just based on the dims names) - and shouldn't we make a point of using geoscience data in our examples? Why are you using `np.fft.fft(()`? We can plot real and imaginary components as well but I'm missing the motivation to be able to write about it. And how did you chose which limits to use for the x-axis? They don't seem consistent across the subplots.

jukent:
@pilotchute test

Issue #345, opened 2023-02-09
`_vertical_remap_extrap` missing vertical dimension name check


anissa111:
Missing warning from internal function `_vertical_remap_extrap` here:
https://github.com/NCAR/geocat-comp/blob/195a0ae13c09c5a103723a415ddee4fb746f213a/src/geocat/comp/interpolation.py#L288

We should be doing a check like exists in `interp_hybrid_to_pressure` that allows them to tell us what the `vertical` name is in their workflow:
https://github.com/NCAR/geocat-comp/blob/195a0ae13c09c5a103723a415ddee4fb746f213a/src/geocat/comp/interpolation.py#L406-L412

This is currently giving a user a bug where another package has a different naming expectation. @hCraker, I believe you wrote this, do you think we could add a quick check here?

anissa111:
Found by @islasimpson

hCraker:
Yes I did write some of the code you've mentioned. After a very quick look, I think a check can be added in case line 288 doesn't produce the expected result. I can look into this week

hCraker:
@islasimpson could you provide me with a short script with the data you're using that produces the error? This will help my diagnose the issue

islasimpson:
Thanks for looking into this.  An example scipt is here /glade/u/home/islas/ERA5interp/isla.ipynb.

hCraker:
HI @islasimpson. I will be releasing a partial fix for the bug you found today. It will be available to installation by tomorrow at the latest. I'm listing is as a partial fix since I'm still getting unexpected values in the stripped down example script you sent me. I have a hunch those values are off because the values for the `lev` coordinate are all 137, but I want to be 100% sure before I close this issue.

islasimpson:
Great.  I'm not seeing any issues with the extrapolation in my version of the code.  I don't see why the lev coordinate all being 137 should matter since the code is using the hybrid coefficients and the surface pressure to determine the pressures and I expect it's not really paying attention to the lev variable, but maybe I'm wrong.  Here is what I'm getting with my version of the code, so it's possible there's just something weird in the data I gave you.  I spent some time trying to figure out what might be going wrong there but I've run out of time to devote to it right now.  But I think it's probably fine to assume that it's ok, given that I'm not having issues with the code I'm currently running.  
![Screenshot at 2023-03-02 11-52-17](https://user-images.githubusercontent.com/41842751/222524313-7d8a0723-7451-4529-879a-7a908d0c5896.png)

Thanks!

[...]

PR #360, opened 2023-03-02
Partial fix for _vertical_remap_extrap bug

hCraker:
## PR Summary

Partially addresses #345 

A user found a bug in `interp_hybrid_to_pressure` when trying to extrapolate values with a non-standard, non-inferable vertical coordinate name. The error was in the helper function `_vertical_remap_extrap`, and this PR resolves the bug. I describe this as a partial fix because the user seems to no longer be having issues, but the minimally verifiable example she sent me is producing some unexpected values. I'm not sure if that is due to some oddities in the input values or if it is another bug in the code.

## PR Checklist

**General**
- [n/a] Link the issue this PR resolves by adding `closes #XXX` to the PR description where XXX is the number of the issue.
- [n/a] Open this PR as a draft if it is not ready for review
- [n/a] Convert this PR from a draft to a full PR before requesting reviewers
- [n/a] Request `@NCAR/geocat` for reviews


jukent (approved):
I'm not sure I 100% understand why naming the vertical pressure coordinate would cause this bug, but if this fix worked - it seems like a good patch until we can uncover more test cases to look at.

hCraker:
Yeah it was a confusing bug at first. It was caused by some spaghetti code I accidentally left in my original PR for the extrapolation feature. I had two variables that were meant to hold the name of the vertical dimension in the helper function, one of which was working incorrectly. I just needed to remove it and use the `lev_dim` variable insteadl

PR #361, opened 2023-03-02
Fix some delta_pressure docs formatting

hCraker:
I'm silly and made a mistake by merging #338 without double checking that the docs formatting was correct. This PR addresses that.

Also @jukent, you mention in the text for the `surface_pressure` argument that it can be a scalar. I added `int` and `float` as accepted classes for that argument. Please let me know if that is incorrect.


hCraker:
Oh boy so RTD and sphinx must have had an update, because a lot of things look strange now.
image

hCraker:
I'm going to merge this despite the changes in the way Sphinx renders the docstrings. It's not breaking anything. Just different. If we decide to change the style later, we can

anissa111 (approved):

Issue #313, opened 2022-12-07
UMBRELLA: Overhaul climatology.py and test_climatology.py


hCraker:
Various issues dealing with `climatology.py` and its test file have been opened. This is an umbrella issue to help keep track of the status of these files.

- [x] #311 
- [x] #265
- [x] #308
- [x] #314
- [x] #198
- [ ] add explicit support for `numpy.ndarray` as inputs

jukent:
[Dev notes for adding numpy support](https://docs.google.com/document/d/1491taNZF5nc01gb07mPQbvDqbMPliQcMBfo1gBjWKSk/edit?usp=sharing)

PR #411, opened 2023-05-11
add custom season method

### LINKED ISSUE #198: Allow user to define seasonal boundaries for climatology_average
Currently `climatology_average` does not allow the user to select the seasonal boundaries. Seasonal climatologies are only calculated for the meteorological seasons, which are defined as follows:
- Winter: December, January, February
- Spring: March, April, May
- Summer: June, July, August
- Fall: September, October, November

This functionality should be added to the climatology module since it is available in NCL's `month_to_season` family of functions.

**Link to original NCL functionality, if applicable**
[month_to_season](https://www.ncl.ucar.edu/Document/Functions/Contributed/month_to_season.shtml)
[month_to_season12](https://www.ncl.ucar.edu/Document/Functions/Contributed/month_to_season12.shtml)
[month_to_seasonN](https://www.ncl.ucar.edu/Document/Functions/Contributed/month_to_seasonN.shtml)

This could be done either by modifying `climatology_average` or by implementing one-to-one copies of NCL's `month_to_season12` and `month_to_seasonN` in Python.

`climatology_average` could be modified to take a parameter defining the first season of interest. This can be done using the NCL convention of three letter strings to represent the months (i.e. JFM for January, February, March as the first season). The function can take this and infer the other three seasons. Then climatologies for all four seasons can be calculated. This would not exactly replicate functionality in NCL's `month_to_season` functions and may be too specialized of a solution. Input from users is needed.

jukent:
## PR Summary
Adds custom season list as kwarg for `climatology_average`
Closes #198

## PR Checklist

**General**
- [x] Add a brief summary of changes to `docs/release-notes.rst`
- [x] Open this PR as a draft if it is not ready for review
- [x] Convert this PR from a draft to a full PR before requesting reviewers
- [x] Request `@NCAR/geocat` for reviews

**Functionality**
- [x] Function is in appropriate module file
- [x] New function(s) intended for public API added to `src/geocat/comp/__init__.py` file

**Testing**
- [x] Tests for function exists in associated module test file
- [x] Tests cover all possible logical paths in your function

**Documentation**
- [x] Docstrings have been added to all new functions ([Documentation Standards](https://geocat.ucar.edu/pages/contributing.html#422-documentation))
- [x] Docstrings have updated with any function changes
- [x] User facing functions have been added to `docs/user_api/index.rst` under their module


jukent:
Target final dataset:
Screen Shot 2023-05-11 at 3 50 57 PM

vs what we currently have:
Screen Shot 2023-05-11 at 3 50 37 PM

I have the values matching up, and am still wrestling with the Xarray dimension(?) name where it says SON instead of season. I'm not sure if we want to add a pandas season index/ if that means something specific to the typical seasons (will research this).

And one issue is that in order to get this values. I have to comment out this block of code borrowed from `month_to_season`:
```
                    # For this season, the last "mean" will be the value for Dec so we drop the last month
                    if season == 'DJF':
                        data_filter = data_filter.isel(
                            {time_dim: slice(None, -1)})
                    # For this season, the first "mean" will be the value for Jan so we drop the first month
                    elif season == 'NDJ':
                        data_filter = data_filter.isel(
                            {time_dim: slice(1, None)})
```
or else that values are off in DJF. I need to look at this math more and know if this is right to not implement (as in are we doing it correctly in the non-custom season version). Any thoughts on that @anissa111

jukent:
Didn't expect this to fail tests in test_climate_anomaly and test_climate_average. Will look into this tomorrow hopefully.

anissa111:
@jukent I'm not sure I know enough about that to have any valuable thoughts right now. I think it could make sense that the two implementations are different enough that this version wouldn't need that block, but I'd want to understand the underlying reason why, does that make sense?

jukent:
Yeah that makes sense. I think when we want winter from one year, we have to deal with it being over the dateline. But when we want winter from all years, it doesn't matter if January is in this winter or the next. Like the grouping of December 22 of January 23 together.

There is still the very first and very last year that could be affected? And that causes the small discrepancy. My hunch is that since we're going through months first it isn't needed or wanted but I'll lay it out better tomorrow.

jukent:
Related to #313

src/geocat/comp/climatologies.py
```diff
@@ -759,20 +795,57 @@ def climatology_average(
...
-        return dset.assign_attrs(attrs)
+        if len(custom_seasons) > 0:  # user specified custom season
+            seasonal_climates = []
+            for season in custom_seasons:
+                try:
+                    months = seasons_pd[season]
+                except KeyError:
+                    raise KeyError(
+                        f"contributed: month_to_season: bad season: SEASON = {season}. Valid seasons include: {list(seasons_pd.keys())}"
+                    )
```
anissa111, line 807: I feel like there's a cleaner way of accomplishing this. Maybe something along the lines of 

```python
if custom_seasons:
    # check for all valid seasons
    if not all([s in season_pd for s in custom_seasons]):
        raise ValueError(...)
```

src/geocat/comp/climatologies.py
```diff
@@ -759,20 +795,57 @@ def climatology_average(
...
+                # Compute the weights for the months in each season so that the
+                # seasonal averages account for months being of different lengths
+                month_length = dset_filter[time_dim].dt.days_in_month
+                weights = month_length / month_length.sum()
+                climatology = (dset_filter * weights).sum(dim=time_dim).values
+
+                seasonal_climates.append(climatology)
+            dset = xr.DataArray(data=seasonal_climates,
+                                dims=["season"],
+                                coords=dict(season=custom_seasons))
```
anissa111, line 830: It seems like we're making a lot of copies here, is there a way to do this more in-place  as well as vectorized to avoid looping here?

anissa111 (changes requested):
I'm a bit concerned with the logic flow as written as mentioned in a few of my comments below for a few reasons, some of which predate this PR

1. I am a little uncozy with the logic structure of 
``` python
if <this setting>:
    do all analysis
    return

if <next setting>:
    do all analysis
    return
```
2. This introduces some edge cases such as what happens if the user specifies custom seasons, but a frequency of anything other than 'season'
3. Apart from the general uncozy mentioned in (1), I think we could be making better use of the `custom_seasons` optional argument here. What if we used a default of `custom_seasons = ['DJF', 'MAM', 'JJA', 'SON']` and then performed the calculation for seasons in the same way instead of separating the season calculation into a "custom" or "default" section? I would think, ideally, the same calculations for custom season should also work for the standard, correct?
4. Seeing the code in this PR, I'm again not totally sure why you've changed some of the inputs to kwargs instead of just arguments. 

Anyway, sorry for the late review, let me know if you'd like a dedicated time for me to explain some of my questions/thinking on suggestions!

src/geocat/comp/climatologies.py
```diff
@@ -684,7 +688,24 @@ def climatology_average(
...
+        - `FMA` : for a season of February, March, and April
+        - `MAM` : for a season of March, April, and May
+        - `AMJ` : for a season of April, May, ad June
+        - `MJJ` : for a season of May, June, and July
+        - `JJA` : for a season of June, July, and August
+        - `JAS` : for a season of July, August, and September
+        - `ASO` : for a season of August, September, and October
+        - `SON` : for a season of September, October, and November
+        - `OND` : for a season of October, November, and December
+        - `NDJ` : for a season of November, December, and January
```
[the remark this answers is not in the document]
jukent: Right now the user would have to enter ['JJA'] - I wanted to type it to accept a string or a list of strings but this seemed easier first.

src/geocat/comp/climatologies.py
```diff
@@ -733,6 +754,21 @@ def climatology_average(
         'season': (None, 'QS-DEC')
     }
 
+    seasons_pd = {
```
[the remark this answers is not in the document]
jukent: Honestly, because it was called that in the month_to_season fx. I can change it to `season_dict` which is more intuitive.

jukent:
2. If the user specifies custom seasons but daily climatoloty, currently the frequency trumps it and the user would get daily climatology. You're right to point that out though. I think that is the behavior we want, but maybe raise a warning to say that their custom seasons will be ignored.
3. Good points. We could do some analysis to see if Xarray's method is faster at all -- but that would work and sounds cleaner for sure.


I'll address your other points and make some edits on Monday.

src/geocat/comp/climatologies.py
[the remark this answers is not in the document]
jukent: just added support for this. We just turn 'JJA' into ['JJA'] before analysis.

src/geocat/comp/climatologies.py
jukent: Tbh this is probably the hardest part for me. My one thought is to make a helper function for what's in the for-loop and then call it using list comprehension. Does that sound like a good approach @anissa111 or do you have another idea?

jukent:
I spent some time trying to solve the error in the climate_anomaly test. The error seems to be in the line `anom = dset.groupby(f"{time_dim}.season") - clim` (line 199) in the climate_anomaly fx. It won't let me take the this difference: "TypeError: unsupported operand type(s) for -: 'int' and 'method'". I tried to trick it by doing groupby on the clim line as well, or groupby and mean or sum (even though there are only one value for each group) and got various error messages. Here `clim` is the method type. I've seen climate anomalies done exactly like this elsewhere online, and it worked before. I think something about artificing the seasonal groupby for `clim` and using Xarray's defaults for `dset` is cauisng the discrepancy. 

I think the easiest solution is to go back to using the default Xarray method when custom seasons aren't specified (instead of specifying them as the default list). I don't love this, but if we're not offering custom season support to climate_anomaly it should work.

Or I could try to add custom_seasons to climate_anomaly, and do a "manual" groupby here as well.

There are other options, but these are what is coming to me right now.

jukent:
Passing local tests!

geocat/comp/climatologies.py
```diff
@@ -280,10 +284,10 @@ def month_to_season(
...
     try:
-        (months, quarter) = seasons_pd[season]
-    except KeyError:
-        raise KeyError(
-            f"contributed: month_to_season: bad season: SEASON = {season}. Valid seasons include: {list(seasons_pd.keys())}"
+        (months, quarter) = seasons_dict[season]
+    except ValueError:
+        raise ValueError(
+            f"contributed: month_to_season: bad season: SEASON = {season}. Valid seasons include: {list(seasons_dict.keys())}"
         )
```
anissa111, line 291: I think it'd be good to propagate this update to the rest of `climatologies.py`

geocat/comp/climatologies.py
```diff
@@ -518,24 +561,54 @@ def climatology_average(
     # Retrieve calendar name
     calendar = _infer_calendar_name(dset[time_dim])
 
+    attrs = {}
+    if keep_attrs or keep_attrs is None:
+        attrs = dset.attrs
+
```
anissa111, line 567:
```suggestion
    attrs = {}
    if keep_attrs in {True, None}:
        attrs = dset.attrs
```
As discussed, we may want to implement a more universal helper function to handle this, but for now, I think this change will reduce unintended results.

geocat/comp/climatologies.py
```diff
@@ -518,24 +561,54 @@ def climatology_average(
...
+                bad_seasons = [
+                    custom_seasons[i]
+                    for i, x in enumerate(valid_season_check)
+                    if not x
+                ]
+                raise ValueError(
+                    f"contributed: climatology_average: bad season(s): {bad_seasons}. Valid seasons include: {list(seasons_dict.keys())}"
+                )
+        else:
+            custom_seasons = ['DJF', 'JJA', 'MAM', 'SON']
```
anissa111, line 584: This is picky, but I think this might actually make more sense the other way around. To me, it seems counterintuitive to proceed with a variable named `custom_seasons`, even if the default seasons are being used.

I would suggest something like:
```python
seasons = ['DJF', 'JJA', 'MAM', 'SON']
if custom_seasons:
    ...
    seasons = custom_seasons
```

geocat/comp/climatologies.py
```diff
@@ -518,24 +561,54 @@ def climatology_average(
...
+        attrs = dset.attrs
+
     if freq == 'season':
-        attrs = {}
-        if keep_attrs or keep_attrs is None:
-            attrs = dset.attrs
-        if xr.infer_freq(dset[time_dim]) != 'MS':
+
+        if custom_seasons:
+            if type(custom_seasons) == str:
```
anissa111, line 571:
```suggestion
            if isinstance(custom_seasons, str):
```

geocat/comp/climatologies.py
```diff
@@ -518,24 +561,54 @@ def climatology_average(
...
+            valid_season_check = [s in seasons_dict for s in custom_seasons]
+            if not all(valid_season_check):
+                bad_seasons = [
+                    custom_seasons[i]
+                    for i, x in enumerate(valid_season_check)
+                    if not x
+                ]
+                raise ValueError(
+                    f"contributed: climatology_average: bad season(s): {bad_seasons}. Valid seasons include: {list(seasons_dict.keys())}"
+                )
```
anissa111, line 582: As you saw this morning, I think we can improve the readability on this particular section. 

```suggestion
            invalid_seasons = [s for s in custom_seasons if s not in seasons_dict]
            if invalid_seasons:
                raise ValueError(
                    f"contributed: climatology_average: bad season(s): {invalid_seasons}. Valid seasons include: {list(seasons_dict.keys())}"
                )
```

test/test_climatologies.py
```diff
@@ -711,7 +711,7 @@ class test_climatology_average(unittest.TestCase):
     day_2_season_clim = xr.Dataset(
         data_vars={'data': (('season', 'lat', 'lon'), season_clim)},
         coords={
-            'season': season_clim_time,
+            'season': np.array(season_clim_time).astype(object),
```
anissa111, line 714: Wait, sorry if I missed something, but why is the trailing `.astype(object)` needed here?

anissa111 (commented):
Mostly just comments from this morning's meeting. I did manage to write something that might be a little more readable for that one section I said I'd take a stab at.

geocat/comp/climatologies.py
jukent: This is much better, thanks

geocat/comp/climatologies.py
jukent: Done. Updated the tests too.

geocat/comp/climatologies.py
jukent: Makes sense to me.

test/test_climatologies.py
jukent: Because otherwise you get type "<U3" as opposed to "object" for each season, which is pretty minor except it makes out output deviate from what it used to be and from what Xarray's is.

jukent: I changed it in line 604 of climatology, and had to change it in the tests too for the result to match.

anissa111: Oh right, okay!

geocat/comp/climatologies.py
```diff
@@ -430,6 +434,7 @@ def calendar_average(
 def climatology_average(
         dset: typing.Union[xr.DataArray, xr.Dataset],
         freq: str,
+        custom_seasons: list = None,
```
anissa111, line 437: Whoops sorry missed this one. The typing should match the docstring.
```suggestion
        custom_seasons: typing.Union[list, str] = None,
```

jukent: Good catch!

anissa111 (approved):
This should be good to go when the tests finish! Good job!!

PR #581, opened 2024-03-20
Add M1 runners to CI

### LINKED ISSUE #546: Add M1 runners to matrix?
GitHub now has M1 runners included in their standard GitHub-hosted runners for Public repositories: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

kafitzgerald:
## PR Summary
Adds M1 runners to CI.  

Closes #546

## PR Checklist

**General**
- [ ] Make an issue if one doesn't already exist
- [ ] Link the issue this PR resolves by adding `closes #XXX` to the PR description where XXX is the number of the issue.
- [ ] Add a brief summary of changes to `docs/release-notes.rst` in a relevant section for the next unreleased release. Possible sections include: Documentation, New Features, Bug Fixes, Internal Changes, Breaking Changes/Deprecated
- [ ] Add appropriate labels to this PR
- [ ] Make your changes in a forked repository rather than directly in this repo
- [ ] Open this PR as a draft if it is not ready for review
- [ ] Convert this PR from a draft to a full PR before requesting reviewers
- [ ] Passes `precommit`. To set up on your local, run `pre-commit install` from the top level of the repository. To manually run pre-commits, use `pre-commit run --all-files` and re-add any changed files before committing again and pushing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest", "macos-latest", "windows-latest" ]
+        os: [ "ubuntu-latest", "macos-latest", "macos-14", "windows-latest" ]
         python-version: ["3.9", "3.10", "3.11", "3.12"]
     steps:
       - name: checkout
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -12,6 +12,7 @@ Upcoming release
 Internal Changes
 ^^^^^^^^^^^^^
 * Additional pre-commit hook for codespell by `Cora Schneck`_ in (:pr:`579`)
+* Add M1 runners to CI by `Katelyn FitzGerald`_ in (:pr:`581`)
 
 v2024.02.0 (February 28, 2024)
 ------------------------------


kafitzgerald:
It looks like there's not a good way to differentiate / specify another macOS platform for Codecov currently ([see options here](https://github.com/codecov/codecov-action/blob/4621ecce093438b398d3d338ab5fe3b038f8e593/src/helpers.ts#L4)).

A few options:
* We could drop the Codecov upload for one set of macOS runners
* We could limit the Codecov uploads even further to a single OS (and even Python version if we wanted) - it doesn't look like you can do much with the additional info anyway at least from the public side of things, but maybe I'm wrong
* We could also drop the macos-latest runs for now or wait for macos-14 to become macos-latest and hold off on testing for M1

@anissa111, thoughts or preferences?

anissa111:
Ahh yeah, we likely don't need to upload codecov for all OS configurations, especially since we're not doing anything different in our module code based on OS. I wonder if there's a way to, like, set a run that we want to upload the codecov info. That's probably outside the scope of this PR though.

For *this* PR, I'd be okay with dropping the upload for either macOS codecov upload if we keep both. I'd also be okay good with dropping `macos-latest`, but we'd need to remember to keep an eye on this in the future. Having both feels a little overkill?

kafitzgerald:
I just went ahead and dropped macos-latest for now, since a) this feels like a more reasonable number of checks w/ better coverage and b) otherwise we'd be solving a problem that's somewhat of an edge case and soon to go away.  Seems like a reasonable balance.  I can log an issue to switch back to macos-latest and keep an eye on that. 

It looks like we have macos-latest checks listed as required though so we'll have to update that temporarily if we move forward.  We'd probably want to update those in some way though regardless.

anissa111:
>It looks like we have macos-latest checks listed as required though so we'll have to update that temporarily if we move forward. We'd probably want to update those in some way though regardless.

I went and switched those around to only require `macos-14`

PR #582, opened 2024-03-22
Reduce DeprecationWarnings in Testing

### LINKED ISSUE #477: Reduce DeprecationWarnings in Testing
Existing tests have a number of warnings being thrown due to `DeprecationWarning`

Would be good to reduce overall number of warnings to future proof testing suites

- Python 3.11
- pip version: 24.0
- conda version: 24.1.0

**Testing context via Python 3.11**
- [ ] [test_climatologies.py](https://github.com/NCAR/geocat-comp/blob/main/test/test_climatologies.py) (`87 passed, 2 warnings in 33.44s`)
- [ ] [test_interpolation.py](https://github.com/NCAR/geocat-comp/blob/main/test/test_interpolation.py) (`25 passed, 19 warnings in 13.97s`)
- [ ] [test_meteorology.py](https://github.com/NCAR/geocat-comp/blob/main/test/test_meteorology.py) (`78 passed, 13 warnings in 10.94s`)
- [ ] [test_gradient.py](https://github.com/NCAR/geocat-comp/blob/main/test/test_gradient.py) (`7 passed, 3 warnings in 10.89s`)
- [ ] [test_stats.py](https://github.com/NCAR/geocat-comp/blob/main/test/test_stats.py) (`24 passed, 21 warnings in 7.83s`)
- [ ] [test_fourier_filters.py](https://github.com/NCAR/geocat-comp/blob/main/test/test_fourier_filters.py) (`17 passed, 2 warnings in 60.29s`)
- [ ] [test_spherical.py](https://github.com/NCAR/geocat-comp/blob/main/test/test_spherical.py) (`6 passed, 2 warnings in 14.23s`)

========================= 244 passed, 46 warnings in 119.93 (0:01:59) ==========================

*link check via CI.yml*

`build finished with problems, 2 warnings.`

cyschneck:
## PR Summary

Closes #477

## PR Checklist

**General**
- [ ] Add a brief summary of changes to `docs/release-notes.rst` in a relevant section for the next unreleased release. Possible sections include: Documentation, New Features, Bug Fixes, Internal Changes, Breaking Changes/Deprecated
- [x] Open this PR as a draft if it is not ready for review
- [ ] Convert this PR from a draft to a full PR before requesting reviewers

### DIFF
--- a/geocat/comp/interpolation.py
+++ b/geocat/comp/interpolation.py
@@ -459,18 +459,17 @@ def interp_hybrid_to_pressure(data: xr.DataArray,
     #     }},
     # )
 
-    # If an unchunked Xarray input is given, chunk it just with its dims
-    if data.chunks is None:
-        data_chunk = dict([
-            (k, v) for (k, v) in zip(list(data.dims), list(data.shape))
-        ])
-        data = data.chunk(data_chunk)
+    # set chunked Xarray input to chunk with dims
+    data_chunk = dict([
+        (k, v) for (k, v) in zip(list(data.dims), list(data.shape))
+    ])
+    data = data.chunk(data_chunk)
 
     # Chunk pressure equal to data's chunks
-    pressure = pressure.chunk(data.chunks)
+    pressure = pressure.chunk(data_chunk)
 
     # Output data structure elements
-    out_chunks = list(data.chunks)
+    out_chunks = list(zip((data_chunk.values())))
     out_chunks[interp_axis] = (new_levels.size,)
     out_chunks = tuple(out_chunks)
     # ''' end of boilerplate


cyschneck:
The filter in `test_dask` is for the warning `RuntimeWarning: invalid value encountered in sqrt` and filtering it is [based on how xarray handles the same issue that dask is triggering](https://github.com/pydata/xarray/issues/8863#issuecomment-2013938201)

The filters in `test_meteorology` are warnings for when `heatindex` is attempting to take the square root of a negative number. Numpy doesn't fail when attempting a square root, but converts the negative number into a `nan`
```
import numpy as np
np.sqrt(np.array([-1, 0, 1]))

<stdin>:1: RuntimeWarning: invalid value encountered in sqrt
array([nan,  0.,  1.])
```
This PR is going to end up leaving two deprecation warnings from `test_stats.py`. These warnings are for [eofuc](https://github.com/NCAR/geocat-comp/blob/e69f3d382ae429702b69bc65a0be5ae11be28fff/geocat/comp/stats.py#L401) and [eofunc_ts](https://github.com/NCAR/geocat-comp/blob/e69f3d382ae429702b69bc65a0be5ae11be28fff/geocat/comp/stats.py#L419). These are both warnings that originate in geocat-comp stats.py that are used for backwards compatibility.

The remaining geocat warning is for `interpolation.py:133` where `UserWarning: Interpolation point out of data bounds encountered return func_interpolate(new_levels, xcoords, data, axis=interp_axis)` when the `func_interpolate` is using `metpy.interpolate.interpolate_1d` function

geocat/comp/interpolation.py
```diff
@@ -459,18 +459,17 @@ def interp_hybrid_to_pressure(data: xr.DataArray,
     #     }},
     # )
 
-    # If an unchunked Xarray input is given, chunk it just with its dims
-    if data.chunks is None:
```
kafitzgerald, line 463: I think we probably want to leave this if statement in here so we're not rechunking data that's already chunked.

It's also somewhat likely this section will change with upcoming work on dask.  Not that we shouldn't address the warning, but just figured it was worth mentioning so you're aware.

test/test_interpolation.py
```diff
@@ -172,6 +172,8 @@ def test_interp_hybrid_to_pressure_extrap_temp(self, temp_in, press_in,
         temp_extrap_expected = ds_out.Tpx.rename(lev_p='plev')
         xr.testing.assert_allclose(temp_extrap_expected, result)
 
+    @pytest.mark.filterwarnings(
```
kafitzgerald, line 175: I'm a little concerned this might be something we should address via better and/or more clear handling in function being tested.  The warning here [seems to indicate](https://unidata.github.io/MetPy/latest/api/generated/metpy.interpolate.interpolate_1d.html#interpolate-1d) that fill values are being added to the output when extrapolation happens.  This doesn't seem to align with the behavior from [what I think is the equivalent function in NCL](https://www.ncl.ucar.edu/Document/Functions/Built-in/vinth2p.shtml) and might be problematic for some users.  

@anissa111 or @erogluorhan, do you happen to know if this was intentional?

It looks like the handling of extrapolation in scipy interpolation functions is more aligned with that from NCL (extrapolates unless a fill value is given)

Fully addressing this is probably beyond the scope of this PR, but I'd be inclined to leave the warning and log an issue unless this was an intentional design decision.

test/test_meteorology.py
```diff
@@ -110,6 +113,9 @@ def test_list_input(self) -> None:
                            self.ncl_gt_1,
                            atol=0.005)
 
+    @pytest.mark.filterwarnings(
```
kafitzgerald, line 116: Unfortunately I'm not super familiar with this function either, but would be curious if this is by design (the nan is a desired output or something we're testing for) or indicating something we should look into / handle better in GeoCAT.

Maybe a question for @anissa111.

kafitzgerald (commented):
I'm a little concerned some of these might also be worth looking into a little more deeply and potentially addressing via better handling or tests in geocat-comp.

cyschneck:
Sounds good, the interpolate function is a lot of interconnected pieced. Let me write up more about how it is populated

cyschneck:
The interpolation warning is generated by the test [`test_interp_hybrid_to_pressure_extrap_geopotential`](https://github.com/NCAR/geocat-comp/blob/47a2ade87c674234240af372f71611d22b19d8f4/test/test_interpolation.py#L175) that is populated from [`press_in`](https://github.com/NCAR/geocat-comp/blob/47a2ade87c674234240af372f71611d22b19d8f4/test/test_interpolation.py#L130) which is populated with the values
```
[[[68619.195 68619.195]
  [69344.43  69290.68 ]
  [70799.516 70643.86 ]]]
```
To be interpolated along the values
```
[ 50000  92500  95000 100000]
```
The test calls `interp_hybrid_to_pressure` as `linear`, so it ends up calling metpy `interpolate_1d`. Specifically, the error being thrown is [because the interpolate values and the coordinates shapes are equal without a fill_value](https://github.com/Unidata/MetPy/blob/37cbf5c74576b15cdfc6ec79fbfde1e33fcd4a37/src/metpy/interpolate/one_dimension.py#L134)

Metpy will raise a warning that:
```
Warn if interpolated values are outside data bounds, will make these the values at end of data range.
```

Since the input values max value `70272.49076537043` is less than max interpolate value `100000` it can't interpolate to the bounds outside the range (without setting a fill_value). So, by default, It adds the values to the [end of the range](https://github.com/Unidata/MetPy/blob/37cbf5c74576b15cdfc6ec79fbfde1e33fcd4a37/src/metpy/interpolate/one_dimension.py#L141)

An easy way to see this is interpolating to 4.5 when the range is to 4:
```
import metpy.interpolate
import numpy as np

x = np.array([1., 2., 3., 4.])
y = np.array([1., 2., 3., 4.])
x_interp = np.array([2.5, 3.5])
metpy.interpolate.interpolate_1d(x_interp, x, y)
```
Where the array position of the max value is out of bounds, so the array appends `nan`
```
metpy.interpolate.interpolate_1d(x_interp, x, y)
<stdin>:1: UserWarning: Interpolation point out of data bounds encountered
array([2.5, nan])
```
When the values are within range, it returns the interpolated values without a nan
```
import metpy.interpolate
import numpy as np

x = np.array([1., 2., 3., 4.])
y = np.array([1., 2., 3., 4.])
x_interp = np.array([2.5, 3.5])
metpy.interpolate.interpolate_1d(x_interp, x, y)
```
```
array([2.5, 3.5])
```

test/test_interpolation.py
cyschneck: Added [some details](https://github.com/NCAR/geocat-comp/pull/582#issuecomment-2037920035) about why this is being generated. Seems like a result of interpolating on a range greater than the data given. It will not fail, but the values above the range are appended as `nan`

kafitzgerald: Yeah, that's what it seemed like.  

In some case this might be ok and expected (after looking more deeply these tests probably fall in that case), but others not.  

It looks like the NCL functions ([vinth2p](https://github.com/NCAR/ncl/blob/develop/ni/src/lib/nfpfort/vinth2p_dp.f) and [vinth2p_ecmwf](https://github.com/NCAR/ncl/blob/develop/ni/src/lib/nfpfort/vinth2p_ecmwf.f)) do something more complex than I thought in practice that might be difficult to replicate though. 

I'd still like to look into this a bit more broadly, but maybe don't have a strong opinion on what we do with the warning for now.

docs/release-notes.rst
```diff
@@ -5,6 +5,14 @@
 Release Notes
 =============
 
+v2024.03.0 (Unreleased)
+---------------------------
+Upcoming release
```
anissa111, line 10:
```suggestion
v2024.04.0 (Unreleased)
-----------------------
Upcoming release
```

test/test_interpolation.py
anissa111: My instinct is to not just filter these warnings in the tests, though I agree that fully addressing this may be outside the scope of this PR.

I also am nervous that this will potentially filter warnings in the future that we may want to know about.

I think I would remove the filter(s) here and log a follow up issue to more thoroughly address this in the code at a later date.

anissa111:
I think I may be a little confused about what's intended to be in this PR. Is this PR supposed to specifically address deprecation warnings only? Also, since this PR will close #477, I'd like to see more of the warnings discussed in that issue addressed here.

anissa111 (changes requested):

test/test_interpolation.py
erogluorhan: Yes, I was going to say something similar. Though Xarray apparently filtered some Dask-resulted warnings, I'd still see filtering of the warnings as a last resort if we couldn't actually overcome the conditions that create such warnings. I see the filtering in `test_interpolate` has now been removed. I am wondering if that in `test_meteorology` is okay or should also be removed.

cyschneck:
In the link checker, there is a redirect happening because the link to the Github Issues tab is not set to check if the user is logged in or not. I can update this for completion sake, but I think the link checker is running into this issue since it is a logged in user.

```
(         contrib: line   46) redirect  https://github.com/NCAR/geocat-comp/issues/new?assignees=&labels=bug%2C+support&projects=&template=bug_report.md&title= - with Found to https://github.com/login?return_to=https%3A%2F%2Fgithub.com%2FNCAR%2Fgeocat-comp%2Fissues%2Fnew%3Fassignees%3D%26labels%3Dbug%252C%2Bsupport%26projects%3D%26template%3Dbug_report.md%26title%3D
(         contrib: line   60) redirect  https://github.com/NCAR/geocat-comp/issues/new?assignees=&labels=enhancement&projects=&template=feature_request.md&title= - with Found to https://github.com/login?return_to=https%3A%2F%2Fgithub.com%2FNCAR%2Fgeocat-comp%2Fissues%2Fnew%3Fassignees%3D%26labels%3Denhancement%26projects%3D%26template%3Dfeature_request.md%26title%3D
```

.github/workflows/pypi.yaml
```diff
@@ -15,10 +15,10 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install setuptools setuptools-scm wheel twine check-manifest
+          python -m pip install build twine check-manifest
```
kafitzgerald, line 18: The changes in here seem reasonable to me, but are they related to the rest of this PR?

I think there may be other build / packaging related changes that could be combined with this as well.  

@anissa111 might have some thoughts too.

docs/release-notes.rst
```diff
@@ -138,7 +146,7 @@ Contributor's Guide to the documentation.
 New Features
 ^^^^^^^^^^^^
 * User-defined seasonal boundaries, ``custom_seasons``, enabled for
-  ``climatology_average`` by `Julia Kent`_ in (:pr:`441`)
+  ``climatology_average`` by `Julia Kent`_ in (:pr:`411`)
```
kafitzgerald, line 149: Good catch on these!

geocat/comp/interpolation.py
```diff
@@ -136,7 +136,7 @@ def _vertical_remap(func_interpolate, new_levels, xcoords, data, interp_axis=0):
 def _temp_extrapolate(data, lev_dim, lev, p_sfc, ps, phi_sfc):
     r"""This helper function extrapolates temperature below ground using the
     ECMWF formulation described in `Vertical Interpolation and Truncation of
-    Model-Coordinate Data <http://dx.doi.org/10.5065/D6HX19NH>`__ by Trenberth,
+    Model-Coordinate Data <https://opensky.ucar.edu/islandora/object/technotes:168>`__ by Trenberth,
```
kafitzgerald, line 139: I'm guessing this is to avoid the redirect, but in practice the DOI is probably the more stable URL.  Kind of a bummer, but I think it's fine.

test/test_dask.py
```diff
@@ -45,6 +45,7 @@ def test_dewtemp_dask(self):
         assert isinstance((out - 273.15).data, dask.array.Array)
         assert np.allclose(out - 273.15, dt_2, atol=0.1)
 
+    @pytest.mark.filterwarnings("ignore:invalid value encountered in sqrt")
```
kafitzgerald, line 48: This just filters the warning out so we don't see it in the test suite, right? 

I don't think the warning is something we or users need to be concerned about (if I recall correctly from the follow on discussion and digging into this a bit), but wonder if filtering it out from the test suite really addresses the broader issue (that users + devs are getting an unnecessary warning).

.github/workflows/pypi.yaml
anissa111: There was a deprecation warning coming out because the packaging folks are trying to encourage people to switch over to `build`.

I tested this out locally and it works essentially the same way. I don't think it'll affect other packaging stuff, but I'll have it on my radar for the next release.

cyschneck: This was a deprecation warning that was being triggered as part of the packaging/documentation work

geocat/comp/interpolation.py
anissa111: Is there a way to ignore redirect warnings for doi links? I'd be open to that, but I don't think it's 100% necessary

cyschneck: The redirect doesn't have to take priority if the DOI is more stable, would it be worth keeping it as a DOI url?

anissa111: The purpose of doi links is to resolve a doi number to the cited resource, so they will always redirect (see https://dx.doi.org/).

.github/workflows/pypi.yaml
kafitzgerald: Nevermind, I think you're right on the packaging stuff.  

Seems great then.

test/test_dask.py
anissa111: Yeah, I agree we shouldn't do this because, at the least, our users will still see it. From some DMs, I actually think this is an unintended behavior bug and I'm looking into it currently. 

In short, in `heat_index`, we apply some adjustments to the values depending on the value of a couple other variables. The adjustment itself should never attempt to square root a negative, but the function (`xr.where`), calculates all the adjustments and then applies them where the filter passes instead of finding where the filter passes and then calculate the adjustment. If that makes sense?

https://github.com/NCAR/geocat-comp/blob/2fbef4e02b430e81de7df192a6bde57cbf62da1a/geocat/comp/meteorology.py#L121-L134

We're doing this partially because of dask compatibility and partially because it's the cleanest way to accept n-dimensional data of numpy/list/xarray types.

kafitzgerald (approved):

anissa111 (approved):

PR #691, opened 2025-02-20
Non nanosecond datetime tests

### LINKED ISSUE #682: Support for non-nanosecond datetime
In short, the broader ecosystem is working to support non-nanosecond datetime objects (see: https://github.com/pydata/xarray/issues/7493) in Pandas 2+ and Xarray's release last month.  This topic was brought to my attention by the recent upstream dev issues w/ Xarray (#681).  

As I was digging through our climatology code today, I noticed we have some things hard coded that will break (in some cases) with these ecosystem changes.

We'll likely want to pin some dependencies and/or make changes and add tests to support this.

Related:
* [Helpful Xarray docs on time coding](https://docs.xarray.dev/en/stable/internals/time-coding.html#internals-timecoding)
* [Xarray docs on non-standard calendars and dates outside the precision range](https://docs.xarray.dev/en/stable/user-guide/weather-climate.html#non-standard-calendars-and-dates-outside-the-precision-range)

kafitzgerald:
## PR Summary
Minor changes to an internal helper function in climatologies borrowed from Xarray in order to support non-nanosecond datetime objects.  The alternative would be to user the corresponding function from Xarray, but it's a bit more complex now (spanning multiple modules) and still private API.  

Thankfully we don't have that much date/time handling code in comp at this point (we mostly rely on xarray and friends) so it should just be this change that's needed.

Added test coverage as well. 

Closes #682

## PR Checklist
**General**

**Testing**

### DIFF
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -11,9 +11,13 @@ Release Notes
 vYYYY.MM.## (unreleased)
 ------------------------
 
+Enhancements
+^^^^^^^^^^^^
+* Add tests and support for non-nanosecond datetime objects by `Katelyn FitzGerald`_ in (:pr:`691`)
+
 Maintenance
 ^^^^^^^^^^^
-* Add support and testing for Python 3.13 by`Katelyn FitzGerald`_ in (:pr:`687`)
+* Add support and testing for Python 3.13 by `Katelyn FitzGerald`_ in (:pr:`687`)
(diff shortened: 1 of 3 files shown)


kafitzgerald:
Link-check is failing because the Pangeo site is temporarily down.

I think we can just ignore for now.

jukent (approved):

anissa111 (approved):

PR #699, opened 2025-03-13
Minimum dependency version testing and backwards compatibility

### LINKED ISSUE #697: Minimum dependency versions / testing
It'd be good to think about a minimum dependency version policy and potentially some associated testing.

Prompted by discussion in #695.

Related:
* [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)
* [SPEC 0](https://scientific-python.org/specs/spec-0000/)
* [Xarray minimum dependency versions](https://docs.xarray.dev/en/stable/getting-started-guide/installing.html#minimum-dependency-versions)

kafitzgerald:
## PR Summary
Adds a minimum dependency environment file and testing to CI and makes a few minor changes for backwards compatibility with earlier versions of Pandas and Xarray to climatology functions and associated testing.  I thought about separating those out into their own PR, but the changes are quite minor.

What's missing:
* documentation for users + devs
* minimum versions in our dependency specification

I'm open to adding these now, but wonder if it might make sense to hold off a bit in case we opt to make changes (e.g. include other packages, change versions, etc.).  Also, I'm not aware of specific limitations to our backward compatibility so these would just be the limits of what we're testing.

Closes #697

## PR Checklist
**General**

### DIFF
--- a/build_envs/min-deps.yml
+++ b/build_envs/min-deps.yml
@@ -10,11 +10,16 @@ dependencies:
   - distributed
   - eofs
   - metpy
+  - netcdf4
   - numpy=1.25
   - packaging
   - pandas=2.0
   - scipy=1.11
   - xarray
+  - xskillscore
   - pip
+  - parameterized
+  - pre-commit
   - pytest
   - pytest-cov
+  - geocat-datafiles
--- a/build_envs/min-deps.yml
+++ b/build_envs/min-deps.yml
@@ -1,0 +1,20 @@
+name: geocat_comp_build
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+  - python=3.10
+  - cf_xarray
+  - cftime
+  - dask-core
+  - distributed
+  - eofs
+  - metpy
+  - numpy=1.25
+  - packaging
+  - pandas=2.0
+  - scipy=1.11
+  - xarray=2023.3
+  - pip
+  - pytest
+  - pytest-cov
--- a/build_envs/min-deps.yml
+++ b/build_envs/min-deps.yml
@@ -14,7 +14,7 @@ dependencies:
   - packaging
   - pandas=2.0
   - scipy=1.11
-  - xarray=2023.3
+  - xarray
   - pip
   - pytest
   - pytest-cov
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
   test:
     # if: |
     #   github.repository == 'NCAR/geocat-comp'
-    name: Python (${{ matrix.python-version }}, ${{ matrix.os }})
+    name: Python ${{ matrix.python-version }}, ${{ matrix.os }}, ${{ matrix.env }}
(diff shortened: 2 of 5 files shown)


kafitzgerald:
I did also rename the checks so we'll have to update those as well.

cyschneck (approved):

build_envs/min-deps.yml
```diff
@@ -0,0 +1,25 @@
...
+  - dask-core
+  - distributed
+  - eofs
+  - metpy
+  - netcdf4
+  - numpy=1.25
+  - packaging
+  - pandas=2.0
+  - scipy=1.11
+  - xarray
```
anissa111, line 18: Just curious, since we wanted to check for specifically xarray backwards compatibility, why not specify a min version here?

anissa111 (approved):
Other than the question I had about specifying an xarray version above, everything here makes sense to me!

I'll go change the required CI names in settings 👍

build_envs/min-deps.yml
kafitzgerald: Good question.

I think I noted this in the issue, but I did try specifying a version for Xarray.  The version recommended by SPEC 0 (2023.3.0 I think) wasn't compatible with Pandas 2 so the environment wouldn't solve.  In hindsight, setting these as max versions in the min deps env and just letting the environment solve would've been a more conservative way to handle this.  It doesn't really align with the broader approach Xarray was taking though (i.e. exactly specifying the min versions and then using those in the dependency specification), but I'm not sure that's important for us.

Anyway, I think there's some good conversation to be had about what min versions we want to use and how much we want to enforce those in our dependency specification.  Acknowledging this, I mostly just made a quick call and went with it.  Not sure it was a good one.

kafitzgerald:
> Other than the question I had about specifying an xarray version above, everything here makes sense to me!
> 
> I'll go change the required CI names in settings 👍
> 
> EDIT: I left the minimum dependency test as optional? I believe that's what we decided on in the dev meeting, but correct me if I'm wrong.

Thanks for doing that!  

And I can't recall offhand what we'd decided, but I don't have a strong opinion either way at this point.

build_envs/min-deps.yml
anissa111: That makes sense to me, I'm good with letting this be our implemented first pass at minimum version testing and adjusting as we go.

kafitzgerald: Alright, in that case I'm going to go ahead an merge.  Definitely not committed to keeping this as is though moving forward.

anissa111:
> > Other than the question I had about specifying an xarray version above, everything here makes sense to me!
> > I'll go change the required CI names in settings 👍
> > EDIT: I left the minimum dependency test as optional? I believe that's what we decided on in the dev meeting, but correct me if I'm wrong.
> 
> Thanks for doing that!
> 
> And I can't recall offhand what we'd decided, but I don't have a strong opinion either way at this point.

I don't either. I think eventually it would make sense to have it be a required test, but I kind of want to see how it works in practice

PR #726, opened 2025-05-08
Delta pressure near surface thickness bugfix

### LINKED ISSUE #725: Correct delta_pressure surface layer thicknesses
As shown/mentioned here: https://github.com/NCAR/geocat-applications/pull/200

kafitzgerald:
## PR Summary
Changes near surface thickness calculations in `delta_pressure` to be consistent with NCL (i.e. based upon the midpoints).

Adds test coverage.

And reminded me we could probably use to optimize this code.  

Closes #725

## PR Checklist
**General**

### DIFF
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -21,6 +21,7 @@ Documentation
 
 Bug Fixes
 ^^^^^^^^^
+* Update near surface pressure thickness calculations in ``delta_pressure`` for consistency with NCL by `Katelyn FitzGerald`_ in (:pr:`726`)
 * Updates internal climatology function to ensure compatibility with both CFTime and Datetime
 indices by `Katelyn FitzGerald`_ in (:pr:`717`)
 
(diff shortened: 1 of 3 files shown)


kafitzgerald:
Linkcheck is failing bc the geocat.ucar.edu website is down.   There's an open PR now to fix that as well.

cyschneck (approved):
Makes sense to me, but where are the equations for the mid-layers being pulled from? Would be good to reference it here and in applications

kafitzgerald:
> Makes sense to me, but where are the equations for the mid-layers being pulled from? Would be good to reference it here and in applications

It's effectively just replicating this logic looking for which layer the surface pressure falls in, which should be relative to the midpoint (via an average) between levels rather than the level specified (how it was before): https://github.com/NCAR/ncl/blob/120dd1a15c1563f923e60605b91ccd566baa28db/ni/src/lib/nfpfort/dpres_plevel_dp.f#L142-L148

I don't know that an equation by itself really helps, since it's just a simple mean and I think the more confusing part is the reference frame.  I can sketch this out next week.  

We'll likely also want to revisit this code to make it a bit more robust and performant, but that's more about the broader functions rather than just this fix.

anissa111 (approved):

kafitzgerald:
I went ahead and logged an issue to remind me to add that equation reference in (#728).

PR #748, opened 2025-08-04
Remove mandatory dask from interpolate functionality

anissa111:
Closes https://github.com/NCAR/geocat-planning/issues/33

This PR:
- removes unnecessary boilerplate code
- introduces branching logic that will only use dask if the user is already using it and other methods are incompatible
- adds a new internal function,  `_interpolate_mb` that re-arranges the order of inputs from map_blocks's expectations to metpy's interpolate expectations
- re-adds pint explicitly to testing requirements as `xr.map_blocks` over metpy's `iterpolation` can occasionally return pint when not intended
- adds testing coverage for new branching logic

TODO:
- ~add back chunk length 1 test~
- ~check test coverage on branching logic~

## PR Checklist
**General**

### DIFF
--- a/geocat/comp/interpolation.py
+++ b/geocat/comp/interpolation.py
@@ -589,7 +589,7 @@ def interp_hybrid_to_pressure(
         # Check again if we got pint back
         if hasattr(output.data, '__module__'):
             if output.data.__module__ == 'pint' and not in_pint:
-                output.data = output.data.to_base_units().magnitude
+                output.data = output.data.to('pascal').magnitude
 
     return output
 


build_envs/environment.yml
```diff
@@ -17,6 +17,7 @@ dependencies:
 # Packages listed below are for testing
   - geocat-datafiles
   - netcdf4
+  - pint
```
kafitzgerald, line 20: I think we should probably add this to the dev group of deps in `pyproject.toml` as well.

And seems fine here if we do use it explicitly in the testing.

jukent (approved):
Thanks for walking through this. Logic is sound.

geocat/comp/interpolation.py
```diff
@@ -476,70 +505,66 @@ def interp_hybrid_to_pressure(
...
+    # if no chunked/dask inputs, just call the function directly
+    else:
+        output = func_interpolate(new_levels, pressure, data.data, axis=interp_axis)
 
     output = xr.DataArray(output, name=data.name, attrs=data.attrs)
 
+    # Check if we've gotten a pint array back from metpy w/o pint in args
+    if hasattr(output.data, '__module__'):
+        if output.data.__module__ == 'pint' and not in_pint:
+            output.data = output.data.to_base_units().magnitude
```
kafitzgerald: I'm probably missing something here, but why the `to_base_units()` here and below?

anissa111: It's to make sure that pint finishes converting everything to the expected output units of pascals before discarding the unit info.

build_envs/environment.yml
anissa111: Ah, yep thanks! Done!

geocat/comp/interpolation.py
kafitzgerald: Do we want to convert explicitly to the intended output units then?

I think the defaults for `to_base_units()` can be configured.   Would something like `to('pascal').magnitude` make more sense?  I'll go ahead and make the suggestion, but I'm honestly not sure I understand this well enough to know if that's a good suggestion or not.

kafitzgerald:
```suggestion
            output.data = output.data.to('pascal').magnitude
```
would this be more robust?  see my earlier comment.

geocat/comp/interpolation.py
```diff
@@ -561,6 +586,11 @@ def interp_hybrid_to_pressure(
             new_levels, lev_dim, data, output, pressure, ps, variable, t_bot, phi_sfc
         )
 
+        # Check again if we got pint back
+        if hasattr(output.data, '__module__'):
+            if output.data.__module__ == 'pint' and not in_pint:
+                output.data = output.data.to_base_units().magnitude
```
kafitzgerald:
```suggestion
                output.data = output.data.to('pascal').magnitude
```

geocat/comp/interpolation.py
anissa111: Hmm. I also don't think I know.

I think your `to('pascal')` is more technically correct, though I don't think it would ever actually make a difference since this check will only be run when metpy has returned pint when not provided with pint inputs. I suppose it might matter if a user has pint configured in their program elsewhere and set a different base unit but did not actually use pint when calling our function? I don't know what the user would prefer/expect in that case, but if our function says it returns pascals, I'd prefer to be certain it does return pascals.

kafitzgerald: Yeah, given that our docs say (Pa) I think it's reasonable to output those and I also would hope this is very much an edge case as you described.  Just makes me a little nervous.

PR #749, opened 2025-08-07
Remove dask as a mandatory dependency

anissa111:
> [!WARNING]
> Intended to be merged after #748 

## PR Summary
This PR removes dask as an explicit dependency and handles the remaining misc dask in geocat comp not covered by #748 

Closes https://github.com/NCAR/geocat-planning/issues/21

## PR Checklist
**General**
- [ ] Add a brief summary of changes to `docs/release-notes.rst` in a relevant section for the upcoming release.

**Testing**

**Documentation**
- [x] Docstrings have been created and/or updated in accordance with [Documentation Standards](https://geocat-comp.readthedocs.io/en/stable/contrib.html#docstrings).

### DIFF
--- a/build_envs/environment.yml
+++ b/build_envs/environment.yml
@@ -5,8 +5,6 @@ dependencies:
   - python>=3.10,<3.14  # minimum support 3.10
   - cf_xarray>=0.3.1    # min version 0.3.1 for dependencies
   - cftime
-  - dask
-  - distributed
   - eofs
   - metpy
   - scipy
--- a/build_envs/min-deps.yml
+++ b/build_envs/min-deps.yml
@@ -6,8 +6,6 @@ dependencies:
   - python=3.10
   - cf_xarray
   - cftime
-  - dask-core
-  - distributed
   - eofs
   - metpy
   - netcdf4
--- a/docs/internal_api/index.rst
+++ b/docs/internal_api/index.rst
@@ -57,12 +57,8 @@ Internal Functionality Helpers
 
    geocat.comp.meteorology._nws_eqn
 
-   geocat.comp.meteorology._relhum
-
    geocat.comp.meteorology._relhum_ice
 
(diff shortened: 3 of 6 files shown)


build_envs/environment.yml
```diff
@@ -5,8 +5,6 @@ dependencies:
   - python>=3.10,<3.14  # minimum support 3.10
   - cf_xarray>=0.3.1    # min version 0.3.1 for dependencies
   - cftime
-  - dask
```
kafitzgerald: Don't we need these for testing though?

build_envs/min-deps.yml
```diff
@@ -6,8 +6,6 @@ dependencies:
   - python=3.10
   - cf_xarray
   - cftime
-  - dask-core
```
kafitzgerald: Same here.

anissa111: I agree given the way we're using this env in testing, but conceptually, should a min-deps environment include testing only dependencies?

build_envs/environment.yml
anissa111: Yeah... I wasn't sure how to handle this in our current environment configurations, but you're right, especially considering that's the way I did it on #748.

Edit update: done!

build_envs/min-deps.yml
kafitzgerald: As configured, it's more "minimum versions" than "minimal dependencies" and we're running the test suite... so for now I think we include them.  Unless I spaced something in my recollection of this (totally possible).

Also, not attached to this approach in the longer term.

anissa111: Ah, yeah. I was assuming it was both, but I see your point and don't really think this semantic thing is worth getting into trying new approaches.

I've put it back like the other one!

erogluorhan (approved):
Looks great; thanks for this!

PR #773, opened 2025-10-09
Zonal mpsi function

### LINKED ISSUE #774: Zonal_MPSI
Feature support for zonal_MPSI analysis of unstructured grid datasets (perhaps expand to work with other datatypes).

Related to https://github.com/NCAR/geocat-planning/issues/11

jukent:
## PR Summary

Zonal_MPSI function for unstructured grid datasets.

Closes #774 

## PR Checklist
**General**

**Functionality**
- [x] New function(s) intended for public API added to `geocat/comp/__init__.py` file

**Testing**

**Documentation**
- [x] Docstrings have been created and/or updated in accordance with [Documentation Standards](https://geocat-comp.readthedocs.io/en/stable/contrib.html#docstrings).
- [x] User facing functions have been added to `docs/user_api/index.rst` under their module

### DIFF
--- a/geocat/comp/meteorology.py
+++ b/geocat/comp/meteorology.py
@@ -1337,12 +1337,12 @@ def delta_pressure(pressure_lev, surface_pressure, pressure_top=None):
 
 def zonal_meridional_psi(
     uxds,
+    lat=(-90, 90, 10),
     meridional_wind_varname=None,
     surface_air_pressure_varname=None,
     plev_coordname=None,
     hyam_coordname=None,
     hybm_coordname=None,
-    lat=(-90, 90, 10),
 ):
     """
     Calculate the zonally averaged meridional streamfunction (mpsi) from a UXarray Dataset.
@@ -1357,6 +1357,11 @@ def zonal_meridional_psi(
             - hyam : Hybrid A coefficients (if meridional_wind is on hybrid sigma-pressure levels)
             - hybm : Hybrid B coefficients (if meridional_wind is on hybrid sigma-pressure levels)
             - uxgrid : Grid information for uxarray
+    lat : tuple, float, or array-like, default=(-90, 90, 10)
+        Latitude specification:
+            - tuple (start, end, step): Computes meridional streamfunction at intervals of `step`.
+            - float: Single latitude
+            - array-like: Latitudes to sample.
     meridional_wind_varname : str, optional
         The variable name for the meridional wind component in `uxds`. If not provided, the function will attempt to find a variable with standard_name 'northward_wind' or long_name 'Meridional wind', or one of the possible names ['V', 'vs', 'meridional_wind'].
     surface_air_pressure_varname : str, optional
@@ -1367,11 +1372,6 @@ def zonal_meridional_psi(
         The coordinate name for hybrid A coefficients in `uxds`. If not provided, the function will attempt to find a coordinate with long_name 'hybrid A coefficient at layer midpoints' or one of the possible names ['hyam', 'hya', 'hybrid_A_midpoints'].
     hybm_coordname : str, optional
         The coordinate name for hybrid B coefficients in `uxds`. If not provided, the function will attempt to find a coordinate with long_name 'hybrid B coefficient at layer midpoints' or one of the possible names ['hybm', 'hyb', 'hybrid_B_midpoints'].
-    lat : tuple, float, or array-like, default=(-90, 90, 10)
-        Latitude specification:
-            - tuple (start, end, step): Computes meridional streamfunction at intervals of `step`.
-            - float: Single latitude
-            - array-like: Latitudes to sample.
     Returns
     -------
     da_mpsi : xarray.DataArray
@@ -1488,10 +1488,10 @@ def zonal_meridional_psi(
         )
     if np.isnan(da_PS_zonal).any():
         raise ValueError(
-            "zonal_meridional_psi: zonal mean of surface air pressure contains all NaN values. "
-            "Cannot compute pressure thickness."
+            "zonal_meridional_psi: zonal mean of surface air pressure for provided latitudes contains NaN values. "
+            "Cannot compute pressure thickness. Check if the `lat` argument matches the range of your data, tried `lat={lat}`."
         )
-    if (da_PS_zonal <= 0).any():
+    if (uxds[surface_air_pressure_varname] <= 0).any():
         raise ValueError("zonal_meridional_psi: surface air pressure must be positive.")
 
     # delta pressure for integration
@@ -1536,21 +1536,21 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
 
 def zonal_mpsi(
     uxds,
+    lat=(-90, 90, 10),
     meridional_wind_varname=None,
     surface_air_pressure_varname=None,
     plev_coordname=None,
     hyam_coordname=None,
     hybm_coordname=None,
-    lat=(-90, 90, 10),
 ):
     return zonal_meridional_psi(
         uxds,
+        lat=lat,
         meridional_wind_varname=meridional_wind_varname,
         surface_air_pressure_varname=surface_air_pressure_varname,
         plev_coordname=plev_coordname,
         hyam_coordname=hyam_coordname,
         hybm_coordname=hybm_coordname,
-        lat=lat,
     )
 
 
--- a/test/test_meteorology.py
+++ b/test/test_meteorology.py
@@ -878,9 +878,28 @@ class Test_zonal_meridional_psi:
 
         with pytest.raises(ValueError) as errinfo:
             zonal_meridional_psi(uxds_bad, lat=self.lat)
-        assert "zonal mean of surface air pressure contains all NaN values" in str(
-            errinfo.value
-        )
+        assert "contains NaN values" in str(errinfo.value)
+
+    def test_zonal_meridional_psi_bad_lat_ranges(self, uxds_plev):
+        # just one outside range of lats in uxds_plev
+        with pytest.raises(ValueError) as errinfo:
+            zonal_meridional_psi(uxds_plev, lat=self.lat.tolist() + [45])
+        assert "contains NaN values" in str(errinfo.value)
+
+        # barely inside range of lats in uxds_plev
+        with pytest.raises(ValueError) as errinfo:
+            zonal_meridional_psi(uxds_plev, lat=[30, 31, 32, 33, 34, 35, 36])
+        assert "contains NaN values" in str(errinfo.value)
+
+        # bigger range of lats outside of uxds_plev
+        with pytest.raises(ValueError) as errinfo:
+            zonal_meridional_psi(uxds_plev, lat=self.lat.tolist() + [45, 46, 47, 48])
+        assert "contains NaN values" in str(errinfo.value)
+
+        # all outside range of lats in uxds_plev
+        with pytest.raises(ValueError) as errinfo:
+            zonal_meridional_psi(uxds_plev, lat=[10, 11, 12, 13, 14, 15])
+        assert "contains NaN values" in str(errinfo.value)
 
     def test_zonal_meridional_psi_custom_varnames(self, uxds_plev):
         """Test providing custom variable names."""


build_envs/environment.yml
```diff
@@ -14,6 +14,7 @@ dependencies:
   - pandas
   - xarray
   - xskillscore>=0.0.17
+  - uxarray>=2025.12.0
```
jukent: I'd say yes. There were updates to UXarray's where and some wrappers that were previously roadblocks to this function. Perhaps conda will naturally resolve the environment to the latest version, but I'd like to let more time pass before trusting that.

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1342,106 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
 
 
 _generate_wrapper_docstring(dpres_plev, delta_pressure)
+
+
+def zonal_mpsi(uxds, lat=list(range(-90, 91, 10))):
```
erogluorhan: A couple questions:

- Is the name "zonal_mpsi" what we want to go forward with? Any alternatives discussed? I recognize finsing something fancy to replace the "mpsi" abbreviation wouldn't be easy though.
- Would it make sense to have `lat` argument default to a tuple of `(start, step, end)` and handle the list creation in the code rather than the func signature? Also, it would take tuple, array-like, or just a scalar, i.e. something similar to [zonal_mean](https://github.com/UXARRAY/uxarray/blob/b8055e50d8cc5f8532c93d53821a12829cf1b258/uxarray/core/dataarray.py#L609)?

build_envs/environment.yml
erogluorhan: Is this pin needed?

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1342,106 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+    - Applies scaling factor based on Earth's radius and gravity.
+    - Currently only supports UXarray Datasets with specified structure.
+    """
+
+    # constants
+    a = 6.371e6  # Earth radius (m)
+    g = 9.80665  # gravity (m/s^2)
+
+    # Basic input validation
+    if not hasattr(uxds, "V") or not hasattr(uxds, "PS"):
```
erogluorhan: I am wondering if this is too limiting, i.e. if there was a meridional wind variable in a dataset without the name (or any attribute) "V", but having the standard_name being "northward_wind", w'd be missing it. Would it be a possible case? Similar applies to "PS".

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1342,106 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+        print("uxds variables:", list(uxds.data_vars))
+        raise AttributeError(
+            "zonal_mpsi: input uxds must contain 'V' and 'PS' variables"
+        )
+
+    if not hasattr(uxds, "uxgrid"):
```
erogluorhan: I get this check, but I don't know if it should be gcomp's concern. Any UXarray dataset has to always have its `uxgrid`, and if not, it is UXarray's major problem I think.

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1342,106 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+        raise AttributeError(
+            "zonal_mpsi: input uxds missing required 'uxgrid' metadata (uxarray)"
+        )
+
+    # Check if interpolation needs to be done
+    if "plev" in uxds["V"].dims:
```
erogluorhan: Similar to above comment on "V" and "PS", maybe we need to look for some CF compliant checks that'd cover any case, e.g. which would use standard name, units, etc?

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1342,106 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+
+    da_dp_zonal = xr.DataArray(
+        np_dp_zonal,
+        dims=["time", "latitudes", "plev"],
+        coords={"plev": da_v_zonal.plev, "latitudes": lat},
+        name="delta_pressure",
+    )
+
+    # check orientation and integrate along pressure dim
+    plev_values = da_v_zonal.coords["plev"].values
```
erogluorhan: I'd highly recommend to avoid `.values` (i.e. stick to `xarray` structures rather than `numpy`) if at all possible to keep things chunks-compatible

geocat/comp/meteorology.py
jukent: 1. We haven't discussed alternatives yet. We aren't bound by character limit so `zonal_meridional_stream()` could work (is including the word function redundant?)
2. Absolutely, I'll make that change right away.

geocat/comp/meteorology.py
jukent: Removed

geocat/comp/meteorology.py
jukent: Fair. I had some checks in but it became very convoluted so I removed them. They are necessary, so I added them back in using 2 new helper functions in `geocat.util` that I imagine could be useful throughout the package. Unfortunately our dev and test data doesn't seem to be cf compliant for meridional wind or surface pressure. I tried using cf_xarray first.

erogluorhan: Would `cf-xarray` be of help on this?

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,116 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+    pressure_names = [
+        'PS',
+        'ps',
+        'surface_pressure',
+        'surface_air_pressure',
+        'pressure_surface',
+    ]
+    hybridA_names = ['hyam', 'hybrid_A_midpoints']
+    hybridB_names = ['hybm', 'hybrid_B_midpoints']
+    plev_names = ['plev', 'pressure_lev', 'pressure_levels']
```
erogluorhan: This possible names-based scheme would probably cover majority of use cases, but I guess I still have some concerns about some easy misses it would have. Also some of these possible name values might not help as expected. 

For instance, 'northward_wind' can mostly show up in `dataarray.attrs.standard_name` rather than being the name of the variable. If the wind variable is simply named "wind" (don't know if data creators would do that though) or anything other than those in this list and has the "standard_name" attribute being "northward_wind", I think `_find_coord()` would miss it.

So, there needs to be some kind of CF-compliant check mechanism here that'd look into not only the var names but also attrs like "standard_name", ~~"units"~~, etc.

EDIT: After follow-ups and careful thinking, checking "units" for detecting a variable like wind might not be reliable.

geocat/comp/gc_util.py
```diff
@@ -27,3 +28,128 @@ def _generate_wrapper_docstring(
...
+                        f"Found {description} '{var_name}' using units attribute only. "
+                        f"This is unreliable - multiple variables may share the same units. "
+                        f"Please verify this is correct and add CF standard_name. {error_parts}",
+                        UserWarning,
+                        stacklevel=3,
+                    )
+                    return var_name
+        error_parts.append(f"Tried units: {units}. ")
+
+    raise KeyError(f"Could not find {description} in dataset. {' '.join(error_parts)}")
```
anissa111, line 137:
```suggestion
    raise KeyError(f"Could not find {description} in dataset. {''.join(error_parts)}")
```

geocat/comp/gc_util.py
```diff
@@ -27,3 +28,128 @@ def _generate_wrapper_docstring(
...
+        error_parts.append(f"Tried long_name: {long_name}. ")
+
+    # Then try direct name match
+    if possible_names:
+        for name in possible_names:
+            if name in ds:
+                return name
+        error_parts.append(f"Tried names: {possible_names}. ")
+
+    # Finally try units match (less reliable)
```
anissa111: Using units here still concerns me. I'm not convinced that units are a good way of determining a match.

geocat/comp/gc_util.py
```diff
@@ -27,3 +28,128 @@ def _generate_wrapper_docstring(
...
+        for var_name in ds.data_vars:
+            var_attrs = ds[var_name].attrs
+            if 'units' in var_attrs:
+                if units in var_attrs.get('units', ''):
+                    warnings.warn(
+                        f"Found {description} '{var_name}' using units attribute only. "
+                        f"This is unreliable - multiple variables may share the same units. "
+                        f"Please verify this is correct and add CF standard_name. {error_parts}",
+                        UserWarning,
+                        stacklevel=3,
```
anissa111: Why `stacklevel=3` here? If my understanding is correct, this would make sense for when `_find_var` is called by `_find_optional_var`, but not when `_find_var` is called directly

test/test_meteorology.py
```diff
@@ -793,3 +796,65 @@ def test_pressure_top_specified(self) -> None:
...
+
+        assert "time" in out.dims
+        assert "latitudes" in out.dims
+        assert "plev" in out.dims
+
+        # shape checks
+        assert out.sizes["latitudes"] == len(self.lat)
+        assert out.sizes["plev"] == len(uxds.V.plev)
+
+        # ---- numerical sanity ----
```
anissa111:
```suggestion
        # ---- numerical coherence ----
```
Two bits:
- small [language best practice](https://it.uw.edu/guides/identity-diversity-inclusion/inclusive-language-guide/#list) note
- I'm not necessarily objecting to having the isinfinite or allclose to 0 asserts, but I don't think I understand why are we doing them? Is there a risk that something will go wrong and we'll get inf/nan/0 values erroneously?

test/test_meteorology.py
```diff
@@ -793,3 +796,65 @@ def test_pressure_top_specified(self) -> None:
...
             pressure_top=pressure_top,
         )
+
+
+# ---- Tests for zonal_meridional_psi ----
+class Test_zonal_meridional_psi:
+    lat = np.arange(36, 45, 1)
+
+    def test_zonal_meridional_psi_pressure_levels(self) -> None:
+        uxds = ux.open_dataset("test/grid_subset.nc", "test/plev_subset.nc")
```
anissa111: This could be a fixture depending on the size/potential for reuse. Also, reading in the files should probably use the same format as the other file readings in our codebase with something like
```python
try:
    uxds = ux.open_dataset("grid_subset.nc", "plev_subset.nc")
except FileNotFoundError:
    uxds = ux.open_dataset("test/grid_subset.nc", "test/plev_subset.nc")
```

test/test_meteorology.py
```diff
@@ -793,3 +796,65 @@ def test_pressure_top_specified(self) -> None:
             self.surface_pressure_scalar,
             pressure_top=pressure_top,
         )
+
+
+# ---- Tests for zonal_meridional_psi ----
```
anissa111: (nitpick, feel free to ignore) we don't use this comment styling anywhere else that I know of in the repo, why here?

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+
+
+def zonal_meridional_psi(
+    uxds,
+    meridional_wind_varname=None,
+    surface_air_pressure_varname=None,
+    plev_coordname=None,
+    hyam_coordname=None,
+    hybm_coordname=None,
+    lat=(-90, 10, 90),
```
anissa111: `(-90, 10, 90)` as the default here is not consistent with the docstring `tuple (start, end, step)` or stated default of `default=(-90, 90, 10)`

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
 _generate_wrapper_docstring(dpres_plev, delta_pressure)
+
+
+def zonal_meridional_psi(
+    uxds,
+    meridional_wind_varname=None,
+    surface_air_pressure_varname=None,
+    plev_coordname=None,
+    hyam_coordname=None,
+    hybm_coordname=None,
```
anissa111, line 1345: Wouldn't it be cleaner (and clearer on the generated documentation) to define the default values here as the ones that will be passed to the util functions to check for?

geocat/comp/gc_util.py
```diff
@@ -27,3 +28,128 @@ def _generate_wrapper_docstring(
...
+    """
+    error_parts = []
+
+    # First try CF standard_name attribute match
+    if standard_name:
+        for var_name in ds.data_vars:
+            var_attrs = ds[var_name].attrs
+            if var_name == standard_name or standard_name in var_attrs.get(
+                'standard_name', ''
+            ):
```
anissa111: (for this and all the following) I feel like we should make this at least a little more robust, like
- checking against the upper and lower case versions of the string
- maybe allowing for some sort of partial match, like "wind" would also match "abc_wind123" (I know that's not a super realistic example) with regex matching

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+        Zonal mean meridional streamf unction, scaled by Earth's geometry and gravity.
+        Dimensions: ["time", "latitudes", "plev"]
+
+    Notes
+    -----
+    - Optionally converts wind data from hybrid sigma-pressure levels to pressure levels.
+    - Computes zonal means and pressure integration deltas.
+    - Integrates over pressure levels, handling orientation.
+    - Applies scaling factor based on Earth's radius and gravity.
+    - Currently only supports UXarray Datasets with specified structure.
```
anissa111, line 1387: I know this might be too big a convo for the comments on PR, but do we really only want to accept uxarray functions here? Why not also at least xarrays?

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+        Latitude specification:
+            - tuple (start, end, step): Computes meridional stream function at intervals of `step`.
+            - float: Single latitude
+            - array-like: Latitudes to sample.
+    Returns
+    -------
+    da_mpsi : xarray.DataArray
+        Zonal mean meridional streamf unction, scaled by Earth's geometry and gravity.
+        Dimensions: ["time", "latitudes", "plev"]
+
```
anissa111, line 1380: Docstring also needs the "See Also" section linking to the corresponding NCL functions

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+                "info": "zonal_meridional_psi: integrated meridional stream function",
+            }
+        )
+    except Exception:
+        pass
+
+    return da_mpsi
+
+
+def zonal_mpsi(
```
anissa111, line 1537: We have an entire utility function for exactly this that generates wrappers and copies the docstrings for ncl-like function names

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+        The coordinate name for hybrid B coefficients in `uxds`. If not provided, the function will attempt to find a coordinate with long_name 'hybrid B coefficient at layer midpoints' or one of the possible names ['hybm', 'hyb', 'hybrid_B_midpoints'].
+    lat : tuple, float, or array-like, default=(-90, 90, 10)
+        Latitude specification:
+            - tuple (start, end, step): Computes meridional stream function at intervals of `step`.
+            - float: Single latitude
+            - array-like: Latitudes to sample.
+    Returns
+    -------
+    da_mpsi : xarray.DataArray
+        Zonal mean meridional streamf unction, scaled by Earth's geometry and gravity.
```
anissa111:
```suggestion
        Zonal mean meridional stream function, scaled by Earth's geometry and gravity.
```

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+    if da_v_zonal.plev.isel(plev=0) < da_v_zonal.plev.isel(plev=-1):
+        integrand = da_v_zonal * da_dp_zonal
+        da_mpsi = integrand.cumsum(dim="plev")
+    else:
+        integrand = (da_v_zonal * da_dp_zonal).isel({"plev": slice(None, None, -1)})
+        da_mpsi = integrand.cumsum(dim="plev").isel({"plev": slice(None, None, -1)})
+
+    # apply scaling factor
+    da_scaling_factor = 2 * np.pi * a * np.cos(lats) / g
+    da_mpsi = da_mpsi * da_scaling_factor
```
anissa111, line 1518: ~~This could be my misunderstanding, but isn't this scaling only valid for certain grid types? Like, specifically only structured ones with equally spaced latitudes?~~

@kafitzgerald, I think I'm remembering something that you flagged on my `nmse` PR,  though completely possible I'm just not familiar enough with the calculation to know what I'm talking about here.

edit: sorry, coming back to say I get that the weighting is for the lat bands, not the unstructured grid itself, but wouldn't this still assume that the given latitudes are evenly spaced?

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+    try:
+        da_mpsi.attrs.update(
+            {
+                "long_name": "zonal mean meridional stream function",
+                "units": "kg/s",
+                "info": "zonal_meridional_psi: integrated meridional stream function",
+            }
+        )
+    except Exception:
+        pass
```
anissa111: (nitpick) bare except makes me a bit nervous here, is there a reason?

Also, don't we create `da_mpsi`? Would updating the attributes ever raise an exception even if we didn't?

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+            possible_names=['hybm', 'hyb', 'hybrid_B_midpoints'],
+            description='coordinate,',
+        )
+    if not plev_coordname:
+        plev_coordname = _find_optional_var(
+            uxds,
+            standard_name='air_pressure',
+            possible_names=['plev', 'pressure_lev', 'pressure_levels'],
+            description='coordinate,',
+        )
```
anissa111, line 1460:
```suggestion
    if not hyam_coordname:
        hyam_coordname = _find_optional_var(
            uxds,
            long_name='hybrid A coefficient at layer midpoints',
            possible_names=['hyam', 'hya', 'hybrid_A_midpoints'],
            description='coordinate',
        )
    if not hybm_coordname:
        hybm_coordname = _find_optional_var(
            uxds,
            long_name='hybrid B coefficient at layer midpoints',
            possible_names=['hybm', 'hyb', 'hybrid_B_midpoints'],
            description='coordinate',
        )
    if not plev_coordname:
        plev_coordname = _find_optional_var(
            uxds,
            standard_name='air_pressure',
            possible_names=['plev', 'pressure_lev', 'pressure_levels'],
            description='coordinate',
        )
```

Given the way `_find_var` is written, say `hya` and `hyb` both have a description of `coordinate` and nothing else, wouldn't this mean that whichever one was found first would become the coord name for both?

Also, I don't have a better suggestion, but is just `'coordinate'` the best default description to look for for all of these?

anissa111 (commented):
Okay, here's my initial review. Also, I'm aware that this is not exactly the place to bring this up, but I feel the need to lodge a comment that I think that geocat-comp ingesting uxarray objects is risky and possibly not in the best interest of geocat-comp.

That being said here's my comments:

- I still feel like we should make uxarray an optional dependency
- `pyproject.toml` and other packaging should be updated to include the optional/required dependency
- uxarray will need to be added to the benchmark environment
- uxarray should also probably be added to upstream testing (as in the upstream test should install uxarray from source in `ci/install-upstream.sh`)
- the helper functions should be added to the internal api documentation
- the helper functions should probably have their own tests
- uxarray specific test(s) for the functions in interpolation that will have uxarray objects passing through them from `zonal_meridional_psi`
- somewhere in the docstring (and also maybe in the code) there should be a citation for the calculation we’re preforming
- at least a handful more tests using inputs with different/missing names/descriptions/etc

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+    ----------
+    uxds : uxarray.UxDataset
+        Input dataset containing the following required fields:
+            - meridional_wind : CF Compliant Meridional wind component (on pressure or hybrid sigma-pressure levels)
+            - surface_air_pressure : CF Compliant Surface pressure
+            - plev : Pressure levels (if meridional_wind is on pressure levels)
+            - hyam : Hybrid A coefficients (if meridional_wind is on hybrid sigma-pressure levels)
+            - hybm : Hybrid B coefficients (if meridional_wind is on hybrid sigma-pressure levels)
+            - uxgrid : Grid information for uxarray
+    meridonal_wind_varname : str, optional
```
anissa111:
```suggestion
    meridional_wind_varname : str, optional
```

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
 
 
 _generate_wrapper_docstring(dpres_plev, delta_pressure)
```
anissa111: This should stay at the bottom of the file (and be used to create the wrapper function made in this PR)

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
...
+    lat=(-90, 10, 90),
+):
+    """
+    Calculate the zonally averaged meridional stream function (mpsi) from a UXarray Dataset.
+
+    Parameters
+    ----------
+    uxds : uxarray.UxDataset
+        Input dataset containing the following required fields:
+            - meridional_wind : CF Compliant Meridional wind component (on pressure or hybrid sigma-pressure levels)
```
anissa111, line 1354: What is the CF complaint name for meridional wind (is there one?)? Is it just `northward_wind`? That seems slightly strange to me.

I'm not finding something explicitly called meridional wind in the CF standard names, though, so maybe

erogluorhan: `northward_wind` is what the `standard_name` attribute should be set to for meridional wind variables in CF complaint files. While the variable name can be anything, there should still be a list of common names data creators tend to use. Hence, a check mechanism for the combination of `standard_name` and common names would be good to cover both CF and non-CF data sets.

geocat/comp/meteorology.py
erogluorhan: I am sure @jukent would already answer this, but I just want to chime in to provide a broader scope: Not only for this function but also for other geocat-comp functions, existing and new, wherever possible, the goal should be to seamlessly support structured grids and unstructured grids through xarray and uxarray, respectively. 

I am hoping to have us work on how to generalize this function (and other functions) to any grids in the next iteration or so.

geocat/comp/interpolation.py
```diff
@@ -383,7 +384,12 @@ def pressure_at_hybrid_levels(psfc, hya, hyb, p0=100000.0):
     """
     # check input types
     in_types = [type(psfc), type(hya), type(hyb)]
-    if not set(in_types).issubset({xr.DataArray, np.ndarray}) or len(set(in_types)) > 1:
+    if (
+        not set(in_types).issubset(
+            {xr.DataArray, np.ndarray, ux.core.dataarray.UxDataArray}
+        )
+        or len(set(in_types)) > 1
+    ):
```
kafitzgerald: Since UXarray inherits from Xarray we may be able to get away with using something like `isinstance()` without needing to rely upon UXarray directly here.  Probably some tradeoffs there though.

And perhaps not an issue if we're committed to UXarray as a required dependency.  However, I also have some concerns about that - mostly in that it's a very heavy dependency and therefore could unnecessarily bloat environments, cause conflicts, etc.

geocat/comp/meteorology.py
```diff
@@ -1340,3 +1343,227 @@ def dpres_plev(pressure_lev, surface_pressure, pressure_top=None):
 
 
 _generate_wrapper_docstring(dpres_plev, delta_pressure)
+
+
+def zonal_meridional_psi(
```
kafitzgerald, line 1338: Did we end up running this name by some folks on the science side of things?

I get that `zonal_mpsi` is not the most descriptive, but am not sure this helps all that much and might be a bit confusing (e.g. psi could mean multiple things and the averaging bit is lost).  I'm not the most up on terminology in this area though so feel free to override this comment.

geocat/comp/meteorology.py
jukent: The description 'coordinate' is not a search criteria, but just a string for the error/warning message. I was thinking about not either not including it or using it more informatively i.e. "meridional wind variable"

geocat/comp/meteorology.py
jukent: @erogluorhan is right, it is `northward_wind`. Not necessarily what I would have chosen if i as in the room making the standards (team `meridional_wind`)

geocat/comp/meteorology.py
jukent: I messaged Brian. I'll let you know. maybe `zonal_mean_meridional_psi()` It's true that `psi` can mean multiple things but streamfunction seems to always be written as 'psi'. Hopefully we find out what scientists expect.
We don't have character limits but `zonal_mean_meridional_streamfunction()` is sooo long that if we get Brian's approval, it would be nice to condense it just a little.

kafitzgerald: Yeah, I agree longer is probably not the way here, but hopefully some additional perspective helps.

geocat/comp/gc_util.py
jukent: I think it needs to be stacklevel=4 when the find optional vars is called, but level 3 when the find mandatory vars i called (one calls the other, both called by zonal_mpsi, then called by the user's code). There might be another way to do this, but if we can remove the less reliable units search, maybe it's a moot point?

geocat/comp/meteorology.py
jukent: Brian likes either `zonal_meridional_psi` or `zonal_mean_meridional_psi`.

geocat/comp/interpolation.py
jukent: `isinstance()` works elsewhere in our type checks. I made the smallest change to the logic that would allow it to pass. I'll reconsider making a different change. @anissa111 did you change this line to specific types for a specific reason?

erogluorhan: >And perhaps not an issue if we're committed to UXarray as a required dependency. However, I also have some concerns about that - mostly in that it's a very heavy dependency and therefore could unnecessarily bloat environments, cause conflicts, etc.

That's a reasonable concern about how heavy this dependency would be. Thinking out loud, maybe we will consider introducing partial UXarray dependency groups, e.g. `uxarray-io`, `uxarray-basic`, etc. which geocat-comp and whoever interested could use (while the default `uxarray` itself would stay as heavy as it is).

geocat/comp/meteorology.py
erogluorhan: I think I can help with this: After UXarray's zonal_mean call on an USG, the result is a list of means on evenly spaced `lat`s. How this is achieved on arbitrarily located faces of USG is described in [this section](https://uxarray.readthedocs.io/en/latest/user-guide/zonal-average.html#step-1-1-understand-the-two-averaging-flavors) of the uxarray user guide (should admit though, a visual depiction would be most helpful). That said, assumption of evenly spaced `lat`s at this point of @jukent 's method seems safe to me.

geocat/comp/interpolation.py
kafitzgerald: > That's a reasonable concern about how heavy this dependency would be. Thinking out loud, maybe we will consider introducing optional UXarray dependency groups, e.g. uxarray-io, uxarray-basic, etc. which geocat-comp and whoever interested could use (while the default uxarray itself would stay as heavy as it is).

Leveraging dependency groups on PyPI and minimizing required dependencies on conda-forge for UXarray could certainly help.  It would require a good bit of work though I suspect.

And in addition to the UXarray viz and I/O deps, we did just drop Numba, Pint, and Dask on geocat-comp not long ago and UXarray would bring those and others along.

Perhaps it's still worth doing as is, but worth some thought IMO.

geocat/comp/meteorology.py
kafitzgerald: Thanks for the context Orhan.

FWIW, I haven't gotten far enough in the weeds on this particular approach to understand the requirements so can't really speak to that.

geocat/comp/interpolation.py
erogluorhan: Extremely helpful to know these; thanks @kafitzgerald ! Will note these for UXarray roadmap.

anissa111: > `isinstance()` works elsewhere in our type checks. I made the smallest change to the logic that would allow it to pass. I'll reconsider making a different change. @anissa111 did you change this line to specific types for a specific reason?

No, when we were working through this bit, this was just the smallest/easiest change to get it to work. I agree with Katelyn that using `isinstance` would be a better call structurally.

jukent:
Remaining work as I understand from this thread (on top of the larger "UXarray as a dependency - is it a good idea?" discussion):

- [ ] make uxarray an optional dependency (`pyproject.toml`?)
- [ ] docstring citation for zonal_mpsi calculation
- [ ] more zonal_mpsi tests using inputs with different/missing names/descriptions/etc
- [ ] interpolation tests for uxarray objects

test/test_meteorology.py
jukent: language updated. yes there is a risk of that. That was one of the main hurdles I was having with generating test data. The function would silently fail by returning all nans.

test/test_meteorology.py
jukent: Made it a fixture!

geocat/comp/meteorology.py
jukent: Oops

geocat/comp/meteorology.py
jukent: I'm not sure. The code is written to check if a variable name kwarg was passed in, if so call it exactly.

If it isn't passed in, use the `_find_vars()` and optional_vars functions. I would have to rewrite that code to treat the user provided string as just another possible name, but I think it should trump all. Happy to discuss more.

geocat/comp/gc_util.py
jukent: I didn't get deep into the regex maxing with added misc numbers, but take a look at the upper, lower, capitalize, all starter words capitalized, and replacing underscores with spaces (vice versa) that I added.

geocat/comp/meteorology.py
jukent: Yes @erogluorhan is right. I'd love to have it accept Xarray datasets as well, but we do call a specific UXarray zonal_mean function that won't work for non-Uxarray Xarray dataset. So it would require I add some extra handling for different datatypes, in the next iteration.

geocat/comp/interpolation.py
jukent: Done

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         env: [""]
         include:
-          - env: "min-deps"
+          - env: "environment"
```
kafitzgerald: I assume this and the change below were because the tests fail without UXarray installed and some confusion over the min-deps.yml environment / purpose (I probably should have documented this more clearly).

Unless I'm missing something, we should revert both changes to this file and instead add UXarray to the min-deps.yml environment.  The purpose of that environment is not so much the minimum number of dependencies, but rather testing to making sure we're backwards compatible with key packages.  

We might also consider specifying the minimum compatible version (presumably the same as from that version pin) so we're testing that as well.

.github/workflows/ci.yml
```diff
@@ -38,8 +38,8 @@ jobs:
         run: |
           echo "TODAY=$(date +%Y-%m-%d)" >> $GITHUB_ENV
 
-          if [[ "${{ matrix.env }}" == "min-deps" ]]; then
-             echo "CONDA_ENV_FILE=build_envs/min-deps.yml" >> $GITHUB_ENV
+          if [[ "${{ matrix.env }}" == "environment" ]]; then
+             echo "CONDA_ENV_FILE=build_envs/environment.yml" >> $GITHUB_ENV
```
kafitzgerald: See comment above.

build_envs/environment.yml
```diff
@@ -20,5 +20,6 @@ dependencies:
   - pint
   - pytest
   - pytest-cov
+  - uxarray>=2025.12.0
```
kafitzgerald: A bit nit-picky here, but I think these are more along the lines of testing packages.  It looks like optional deps (e.g. dask) are in the category above.

ci/install-upstream.sh
```diff
@@ -26,7 +27,8 @@ python -m pip install \
     numpy \
     pandas \
     scipy \
-    xarray
+    xarray \
+    uxarray
```
kafitzgerald: I don't think UXarray uploads nightly wheels so we'll want to install it from GitHub like xskillscore, MetPy, etc.

docs/release-notes.rst
```diff
@@ -8,6 +8,17 @@
 Release Notes
 =============
 
+
+vYYYY.MM.## (unreleased)
+------------------------
+This release has updates to our internal structure to ingest UXarray datasets in a new zonal meridional stream function for unstructured grid datafiles using UXarray in ``meteorology.zonal_meridional_psi``. There are also two new helper functioins ``gc_util._find_coord`` and ``gc_util._find_optional_coord``
```
kafitzgerald:
```suggestion
This release has updates to our internal structure to ingest UXarray datasets in a new zonal mean meridional stream function for data on unstructured grids using UXarray in ``meteorology.zonal_meridional_psi``. There are also two new helper functions ``gc_util._find_coord`` and ``gc_util._find_optional_coord``
```

docs/release-notes.rst
```diff
@@ -8,6 +8,17 @@
...
+
+New Features
+^^^^^^^^^^^^
+* Zonal Mean Meridional Stream Function for unstructured grid datafiles using
```
kafitzgerald:
```suggestion
* Zonal mean meridional stream function for data on unstructured grids using
```

kafitzgerald (commented):
A few notes on the dependency specification.

I didn't look over the new functionality or tests, but can do that next week if helpful.

.github/workflows/ci.yml
anissa111: +1 here

geocat/comp/meteorology.py
```diff
@@ -1334,9 +1342,218 @@ def delta_pressure(pressure_lev, surface_pressure, pressure_top=None):
...
+    pages B-17 to B-18.
+    https://opensky.ucar.edu/islandora/object/technotes%3A151
+
+    See Also
+    --------
+    Related NCL Functions:
+    `zonal_mpsi <https://www.ncl.ucar.edu/Document/Functions/Built-in/zonal_mpsi.shtml>`__
+    """
+
+    if has_uxarray:
```
anissa111: Probably a nitpick, but instead of nesting the entire rest of the function inside the if, I think it's way more clear to do 
```python
if not has_uxarray:
	# error

...
```

geocat/comp/meteorology.py
```diff
@@ -1334,9 +1342,218 @@ def delta_pressure(pressure_lev, surface_pressure, pressure_top=None):
...
+            - float: Single latitude
+            - array-like: Latitudes to sample.
+    Returns
+    -------
+    da_mpsi : xarray.DataArray
+        Zonal mean meridional streamfunction, scaled by Earth's geometry and gravity.
+        Dimensions: ["time", "latitudes", "plev"]
+
+    Notes and Correspondence:
+    -----
```
anissa111: Currently, this looks funny on the generated docs:

Image


```suggestion
    Notes
    -----
```

geocat/comp/meteorology.py
```diff
@@ -1334,9 +1342,218 @@ def delta_pressure(pressure_lev, surface_pressure, pressure_top=None):
...
+    See Also
+    --------
+    Related NCL Functions:
+    `zonal_mpsi <https://www.ncl.ucar.edu/Document/Functions/Built-in/zonal_mpsi.shtml>`__
+    """
+
+    if has_uxarray:
+        # constants
+        a = 6.371e6  # Earth radius (m)
+        g = 9.80665  # gravity (m/s^2)
```
anissa111: Not sure how big an issue this really is, but not all the models necessarily use the same value for earth's radius.

It seems like metpy uses `6371008.7714` by default. 

If nothing else, we should specify in the docstring what we're using for these values.

docs/user_api/index.rst
anissa111, line 105: Wrapper function should also be added to generated docs

```suggestion
   meteorology.dpres_plev
   meteorology.zonal_mpsi
```

geocat/comp/meteorology.py
```diff
@@ -1334,9 +1342,218 @@ def delta_pressure(pressure_lev, surface_pressure, pressure_top=None):
...
+):
+    return zonal_meridional_psi(
+        uxds,
+        meridional_wind_varname=None,
+        surface_air_pressure_varname=None,
+        plev_coordname=None,
+        hyam_coordname=None,
+        hybm_coordname=None,
+        lat=(-90, 90, 10),
+    )
```
anissa111, line 1554: Wrapper function should pass along any optional inputs
```suggestion
def zonal_mpsi(
    uxds,
    meridional_wind_varname=None,
    surface_air_pressure_varname=None,
    plev_coordname=None,
    hyam_coordname=None,
    hybm_coordname=None,
    lat=(-90, 90, 10),
):
    return zonal_meridional_psi(
        uxds,
        meridional_wind_varname=meridional_wind_varname,
        surface_air_pressure_varname=surface_air_pressure_varname,
        plev_coordname=plev_coordname,
        hyam_coordname=hyam_coordname,
        hybm_coordname=hybm_coordname,
        lat=lat,
    )
```

.github/workflows/ci.yml
jukent: Okay I reversed it. I was confused about it and still kind of am. I completely removed UXarray from `environment.yml` please correct me if that wasn't the move.

geocat/comp/meteorology.py
jukent: The a value I used was taken from the NCL code, maybe it's best to use the MetPy default?

geocat/comp/meteorology.py
```diff
@@ -3,7 +3,15 @@
 import warnings
 import xarray as xr
 
-from .gc_util import _generate_wrapper_docstring
+from .gc_util import _generate_wrapper_docstring, _find_var, _find_optional_var
+from .interpolation import interp_hybrid_to_pressure
+
+try:
+    import uxarray as ux
```
erogluorhan: For a true optional dependency experience for the time being, this import can be moved into the zonal mpsi function (it also helps with the import times of the rest as [pointed out](https://github.com/NCAR/geocat-comp/pull/773#issuecomment-3962451536) by @anissa111 )

docs/release-notes.rst
```diff
@@ -8,6 +8,17 @@
...
 
+
+vYYYY.MM.## (unreleased)
+------------------------
+This release has updates to our internal structure to ingest UXarray datasets in a new zonal mean meridional stream function for data on unstructured grids using UXarray in ``meteorology.zonal_meridional_psi``. There are also two new helper functions ``gc_util._find_coord`` and ``gc_util._find_optional_coord``
+
+New Features
+^^^^^^^^^^^^
+* Zonal mean meridional stream function for data on unstructured grids using
+UXarray in ``meteorology.zonal_meridional_psi`` by `Julia Kent`_ in (:pr:`773`)
```
anissa111: Unfortunately, we can't make line breaks in the bullet points, it does this in the generated docs:

Image

```suggestion
New Features
^^^^^^^^^^^^
* Zonal mean meridional stream function for data on unstructured grids using UXarray in ``meteorology.zonal_meridional_psi`` by `Julia Kent`_ in (:pr:`773`)
```

docs/release-notes.rst
```diff
@@ -8,6 +8,17 @@
 Release Notes
 =============
 
+
+vYYYY.MM.## (unreleased)
+------------------------
+This release has updates to our internal structure to ingest UXarray datasets in a new zonal mean meridional stream function for data on unstructured grids using UXarray in ``meteorology.zonal_meridional_psi``. There are also two new helper functions ``gc_util._find_coord`` and ``gc_util._find_optional_coord``
```
anissa111: (nitpick, feel free to ignore) I don't think we need to include the internal helper funcs in the pubic release notes

```suggestion
This release has updates to our internal structure to ingest UXarray datasets in a new zonal mean meridional stream function for data on unstructured grids using UXarray in ``meteorology.zonal_meridional_psi``.
```

test/test_meteorology.py
```diff
@@ -793,3 +796,214 @@ def test_pressure_top_specified(self) -> None:
...
+
+        xr.testing.assert_allclose(out_func, out_manual)
+
+    @pytest.mark.xfail(reason="Dataset may not cover extreme latitudes")
+    def test_zonal_meridional_psi_raises_on_nan_surface_pressure(self, uxds_plev):
+        """Ensure zonal_meridional_psi fails when surface pressure contains NaNs."""
+        uxds_bad = uxds_plev.copy()
+        uxds_bad["PS"][:] = np.nan
+
+        zonal_meridional_psi(uxds_bad, lat=self.lat)
```
anissa111: This isn't how xfails are supposed to be used. xfails are, from my understanding, supposed to be tests that *should* pass, but for some known reason, do not, like, something not being fully implemented yet. By just marking the entire test as an xfail, it will "pass" the test if anything in it fails for any reason. Also, not sure what this test has to do w/ extreme lats.

If you're trying to check for a specific error, this would be appropriate:

```python
def test_zonal_meridional_psi_raises_on_nan_surface_pressure(self, uxds_plev):
    """Ensure zonal_meridional_psi fails when surface pressure contains NaNs."""
    uxds_bad = uxds_plev.copy()
    uxds_bad["PS"][:] = np.nan

    with pytest.raises(ValueError) as errinfo:
        zonal_meridional_psi(uxds_bad, lat=self.lat)
    assert "surface air pressure contains NaN" in str(errinfo.value)
```

However (was trying this because of the comment on the next test) I don't think this is necessary because if you just sprinkle nans instead of turning all the values to nans, the function works fine ⬇️

```python
def test_zonal_meridional_psi_works_with_nans(self, uxds_plev):
    """Ensure zonal_meridional_psi fails when surface pressure contains NaNs."""
    uxds_bad = uxds_plev.copy()

    # sprinkle nans
    uxds_bad["PS"][0][3] = np.nan
    uxds_bad["PS"][0][80] = np.nan
    uxds_bad["PS"][0][15] = np.nan

    zonal_meridional_psi(uxds_bad, lat=self.lat)
```

test/test_meteorology.py
```diff
@@ -793,3 +796,214 @@ def test_pressure_top_specified(self) -> None:
...
+
+    @pytest.mark.xfail(reason="Dataset may not cover extreme latitudes")
+    def test_zonal_meridional_psi_extreme_latitudes(self, uxds_plev):
+        """Test with latitude values at poles."""
+        out = zonal_meridional_psi(uxds_plev, lat=np.array([-90, 0, 90]))
+        assert np.isfinite(out).all()
```
anissa111: Following on to the xfails will "pass" a test that fails for any reason, the actual error this test throws is:
```python
ValueError: zonal_meridional_psi: surface air pressure contains NaN values. Cannot compute pressure thickness.
```

Which is, I assume, not what you were trying to test for here because `uxds_plev` does not have NaN values and the xfail was covering up the bug.

I assume that if you want to test the function with points at/near the poles (which sounds like something we should do), you'll need a dataset that actually has points at the poles to test.

```python
>>> min(uxds_plev.uxgrid.node_lat.values)
np.float64(35.06064797646591)

>>> max(uxds_plev.uxgrid.node_lat.values)
np.float64(44.93780349248086)
```

What this test was actually finding was that this line in `zonal_meridional_psi` returns nan values because the lats provided are out of range for the data, not that the user provided surface pressure with nans. 
```python
da_PS_zonal = uxds[surface_air_pressure_varname].zonal_mean(lat=lat)
```

From this, I think we should
- either remove this test or get a dataset that we could test close to the poles
- make the error in `meteorology.py` reflect that the `zonal_mean` returned nan values, not that the user provided nan values.
- if it makes sense to, make a separate check to see if the user provided nan surface pressure values

test/test_meteorology.py
```diff
@@ -793,3 +796,214 @@ def test_pressure_top_specified(self) -> None:
...
+
+    def test_zonal_meridional_psi_cos_latitude_weighting(self, uxds_plev):
+        """Test that cosine latitude weighting is applied."""
+        out_low = zonal_meridional_psi(uxds_plev, lat=np.array([36]))
+        out_high = zonal_meridional_psi(uxds_plev, lat=np.array([44]))
+
+        assert np.isfinite(out_low).all()
+        assert np.isfinite(out_high).all()
```
anissa111: This test does not actually tests if the cos lat weights are applied, it just checks to see if the results are finite, which has nothing to do with the weighting.

test/test_meteorology.py
```diff
@@ -793,3 +796,214 @@ def test_pressure_top_specified(self) -> None:
...
+
+        assert isinstance(out, xr.DataArray)
+        assert out.sizes["latitudes"] == len(self.lat)
+
+    def test_zonal_meridional_psi_single_latitude(self, uxds_plev):
+        """Test with single latitude value."""
+        out = zonal_meridional_psi(uxds_plev, lat=40.0)
+
+        assert isinstance(out, xr.DataArray)
+        assert "latitudes" in out.dims
```
anissa111: Similar to above, this test does not test what it says it's testing. This (or similar) would be closer:

```suggestion
    def test_zonal_meridional_psi_single_latitude(self, uxds_plev):
        """Test with single latitude value."""
        out = zonal_meridional_psi(uxds_plev, lat=40.0)

        assert isinstance(out, xr.DataArray)
        assert out.latitudes == 40
```

test/test_meteorology.py
```diff
@@ -793,3 +796,214 @@ def test_pressure_top_specified(self) -> None:
...
+
+        # Check that integration handled orientation correctly
+        assert np.isfinite(out).all()
+
+    def test_zonal_meridional_psi_descending_pressure(self, uxds_hybrid):
+        """Test with descending pressure levels (typical for hybrid coords)."""
+        out = zonal_meridional_psi(uxds_hybrid, lat=self.lat)
+
+        # Check that integration handled orientation correctly
+        assert np.isfinite(out).all()
```
anissa111: Again, neither of these tests test what they say they're testing.

Since `uxds_plev` and `uxds_hybrid` use the same pressure levels (once converted from hybrid), the directional difference could be actually tested by comparing them in a single test:


```suggestion
def test_zonal_meridional_psi_descending_vs_ascending(self, uxds_hybrid, uxds_plev):
    # descending hybrid coords
    out_hybrid = zonal_meridional_psi(uxds_hybrid, lat=self.lat)
    
    # ascending (matching) plev coords
    out_plev = zonal_meridional_psi(uxds_plev, lat=self.lat)

    # Check that integration handled orientation correctly
    np.testing.assert_allclose(out_hybrid, out_plev)
```

test/test_meteorology.py
```diff
@@ -793,3 +796,214 @@ def test_pressure_top_specified(self) -> None:
...
+
+    def test_zonal_meridional_psi_scaling_magnitude(self, uxds_plev):
+        """Test that Earth geometry scaling produces reasonable magnitudes."""
+        out = zonal_meridional_psi(uxds_plev, lat=self.lat)
+
+        # Values should be large due to Earth radius scaling (2*pi*a/g term)
+        # Typical values are O(10^10) kg/s
+        assert np.abs(out.values).max() > 1e6
+
```
anissa111: Not needed, arguably implicitly tested in other correctness tests
```suggestion
```

test/test_meteorology.py
```diff
@@ -793,3 +796,214 @@ def test_pressure_top_specified(self) -> None:
...
+    def test_zonal_meridional_psi_case_insensitive_finding(self, uxds_plev):
+        """Test that variable finding works with case variations."""
+        # This tests the _find_var functionality indirectly
+        out = zonal_meridional_psi(uxds_plev, lat=self.lat)
+        assert isinstance(out, xr.DataArray)
```
anissa111: Does not test what it says its testing.

test/test_meteorology.py
```diff
@@ -793,3 +796,214 @@ def test_pressure_top_specified(self) -> None:
...
+
+    def test_zonal_meridional_psi_pressure_integration_correctness(self, uxds_plev):
+        """Test that pressure integration produces monotonic results."""
+        out = zonal_meridional_psi(uxds_plev, lat=self.lat)
+
+        # Stream function should vary smoothly with pressure
+        # (not necessarily monotonic, but should have structure)
+        assert not np.all(np.diff(out.values, axis=-1) == 0)
```
anissa111: The name of the test does not match what it says it's testing, nor does it actually test if it is monotonic. It specifically asserts that not all the differences are 0, which definitely does not test for monotonicity. I'm not sure that this is a meaningful assert at all.


Not sure what to replace this with (if anything), but here's what the np.diff looks like for an arbitrary latitude:

```python
np.diff(out.sel(latitudes=40).values)
Out[8]: 
array([[ 0.00000000e+00,  0.00000000e+00,  0.00000000e+00,
         4.69499918e+10,  1.15153108e+11,  1.42510024e+11,
         1.50124695e+11,  1.16205181e+11,  1.15151133e+11,
         8.87519908e+10,  3.97662401e+10,  2.09608124e+09,
        -9.48103900e+09, -2.19168168e+10, -2.17866127e+10,
        -1.98191259e+10, -8.11905879e+09, -6.43628423e+09,
         0.00000000e+00,  0.00000000e+00]])
```

test/test_meteorology.py
```diff
@@ -793,3 +796,214 @@ def test_pressure_top_specified(self) -> None:
...
+
+        assert "time" in out.dims
+        assert out.sizes["time"] == uxds_plev.sizes["time"]
+
+    def test_zonal_meridional_psi_hybrid_creates_plev_coord(self, uxds_hybrid):
+        """Test that hybrid interpolation creates 'plev' coordinate."""
+        out = zonal_meridional_psi(uxds_hybrid, lat=self.lat)
+
+        assert "plev" in out.dims
+        assert "plev" in out.coords
```
anissa111: Arguably duplicates coverage with `test_zonal_meridional_psi_dimension_order`, which tests for `time` and `plev` existence in `out.dims` and `test_zonal_meridional_psi_pressure_levels`, which also specifically asserts that `"time" in out.dims`, `"latitudes" in out.dims` and `"plev" in out.dims`.

```suggestion
```

test/test_gc_util.py
```diff
@@ -0,0 +1,251 @@
...
+        assert result == 'TEMP'
+
+    def test_priority_long_name_over_possible_names(self, sample_dataset):
+        """Test that long_name has priority over possible_names."""
+        result = _find_var(
+            sample_dataset,
+            long_name='Air Temperature',
+            possible_names=['pressure', 'TEMP'],
+        )
+        assert result == 'TEMP'
```
anissa111: Do not test what they say they're testing, cannot test priority order if both could find the variable.

```suggestion
    def test_priority_standard_name_over_possible_names(self, sample_dataset):
        """Test that standard_name has priority over possible_names."""
        result = _find_var(
            sample_dataset,
            standard_name='air_temperature',
            possible_names=['pressure'],
        )
        assert result == 'TEMP'

    def test_priority_long_name_over_possible_names(self, sample_dataset):
        """Test that long_name has priority over possible_names."""
        result = _find_var(
            sample_dataset,
            long_name='surface pressure',
            possible_names=['TEMP'],
        )
        assert result == 'pressure'
```

test/test_gc_util.py
```diff
@@ -0,0 +1,251 @@
...
+                standard_name='not_here',
+                long_name='also_not_here',
+                possible_names=['nope'],
+            )
+
+        error_msg = str(excinfo.value)
+        assert 'Tried standard_name:' in error_msg
+        assert 'Tried long_name:' in error_msg
+        assert 'Tried names:' in error_msg
+
```
anissa111: Arguably duplicate of `test_not_found_raises_keyerror`, which tests for all components in error message

test/test_gc_util.py
```diff
@@ -0,0 +1,251 @@
...
+            long_name='does_not_exist',
+            possible_names=['nope', 'also_nope'],
+        )
+        assert result is None
+
+    def test_none_not_exception(self, sample_dataset):
+        """Test that None is returned, not an exception object."""
+        result = _find_optional_var(sample_dataset, standard_name='nonexistent')
+        assert result is None
+        assert not isinstance(result, Exception)
```
anissa111: (nitpick) if `result` is an exception, the test would have already failed before this assert

```suggestion
    def test_none_not_exception(self, sample_dataset):
        """Test that None is returned, not an exception object."""
        result = _find_optional_var(sample_dataset, standard_name='nonexistent')
        assert result is None
```

docs/release-notes.rst
anissa111: (not sure why the suggested diff looks like that, I just removed the line break)

test/test_meteorology.py
jukent: Hmm I only care if they're all nans. I'll update the test to reflect this

test/test_meteorology.py
jukent: Are we sure that our hybrid is descending our plev is ascending? I'll double check

jukent: So hybrid has levels that are going up in value and plev has values that are going down in value (so ascending the atmosphere) so this is correct.

test/test_meteorology.py
jukent: I think this isn't needed with the tests in test_util for that helper function

test/test_gc_util.py
jukent: Removed

test/test_gc_util.py
jukent: Incorporated

test/test_gc_util.py
jukent: Incorporated

test/test_meteorology.py
jukent: Removed

test/test_meteorology.py
jukent: Removed

test/test_meteorology.py
jukent: Removed

test/test_meteorology.py
jukent: Removed

test/test_meteorology.py
jukent: Incorporated

test/test_meteorology.py
jukent: Incorporated

test/test_meteorology.py
jukent: Removed, not sure how to test effectively

test/test_meteorology.py
jukent: Removed all xfail

test/test_meteorology.py
anissa111: But what I'm trying to say here [and in the next comment](https://github.com/NCAR/geocat-comp/pull/773#discussion_r2875059093) is that whether the user provided surface air pressure contains some/all nans or not is not what is actually being checked. 

The way it is currently written, a user could provide no nans and still get this error saying they provided all nans because what's actually being checked in the if block that triggers this error message is not the provided surface air pressure, it's the calculated zonal mean of the provided surface air pressure.

from meteorology.py:
```python
da_PS_zonal = uxds[surface_air_pressure_varname].zonal_mean(lat=lat)
...
if np.isnan(da_PS_zonal).any():
    raise ValueError(
        "zonal_meridional_psi: surface air pressure contains all NaN values. "
        "Cannot compute pressure thickness."
    )
```

Here's two other examples that currently trigger the "all nans" error that don't have all nans:
```python
# include some lats not covered in the original dataset (no nans in input at all)
zonal_meridional_psi(uxds_plev, lat=self.lat.tolist() + [45, 56])
```
```python
# for some reason, a certain lat band is just nans, the rest of the data is not nans

uxds_bad = uxds_plev.copy()

# get indexes for faces `39 < face_lat < 41`
inrange = set(np.where(uxds_bad["PS"].uxgrid.face_lat.values > 39)[0].tolist()).intersection(
    set(np.where(uxds_bad["PS"].uxgrid.face_lat.values < 41)[0].tolist()))
# set those indexes to nans
for i in inrange:
    uxds_bad["PS"][0][i] = np.nan

zonal_meridional_psi(uxds_bad, lat=self.lat)
```

jukent: Thanks @anissa111 I'll have to think about this some more.

jukent: Okay updating the original error message to: "zonal_meridional_psi: zonal mean of surface air pressure contains all NaN values. "

and the test to:
```
    def test_zonal_meridional_psi_raises_on_nan_zonal_mean(self, uxds_plev):
        """Test error when zonal mean contains NaN (achieved by passing all-NaN input)."""
        uxds_bad = uxds_plev.copy()
        uxds_bad["PS"][:] = np.nan

        with pytest.raises(ValueError) as errinfo:
            zonal_meridional_psi(uxds_bad, lat=self.lat)
        assert "zonal mean of surface air pressure contains all NaN values" in str(
            errinfo.value
        )
   ```

Comment (anissa111) [reply test/test_meteorology.py]: I do think that the error message is a bit better now, but you haven't actually resolve the problems shown in my two test examples.

There's a couple things going on here.
1. You're checking `.any()`, but the error message implies you're checking `.all()`.
2. Whether or not `.zonal_mean` returns nans seems to depend on whether or not the latitudes provided have data in them to take the mean of. This is primarily what I was getting at with my previous comment. The error message is not helpful for this.
3. We're not explicitly checking whether the lats provided are valid lats for the dataset. What do we want to happen if the user provides either data that has nans for entire lat ranges or lats ranges outside of the provided data?

If we change the `.any()`s to `.all()`s in `meteorology.py`, the rest of the function handles my two test cases differently. For:
```python
# include some lats not covered in the original dataset (no nans in input at all)
zonal_meridional_psi(uxds_plev, lat=self.lat.tolist() + [45, 56])
```
We get:
```python
# Determine which layers to calculate thickness for based upon midpoints
midpoints = np.array((pressure_lev + np.roll(pressure_lev, shift=1)) / 2.0)
midpoints[0] = 0.0
[indices] = np.nonzero(midpoints < surface_pressure)

>       start_level = min(indices)
              		  ^^^^^^^^^^^^
E       ValueError: min() iterable argument is empty
../geocat/comp/meteorology.py:1201: ValueError
```

For:
```python
uxds_bad = uxds_plev.copy()

# get indexes for faces `39 < face_lat < 41`
inrange = set(np.where(uxds_bad["PS"].uxgrid.face_lat.values > 39)[0].tolist()).intersection(
    set(np.where(uxds_bad["PS"].uxgrid.face_lat.values < 41)[0].tolist()))
# set those indexes to nans
for i in inrange:
    uxds_bad["PS"][0][i] = np.nan

zonal_meridional_psi(uxds_bad, lat=self.lat)
```
We get a return value, but I have no idea how valid it is:
```python
out.sel(latitudes=40).values
Out[4]: 
array([[-7.29149507e+11, -7.29149507e+11, -7.29149507e+11,
        -7.29149507e+11, -6.82199516e+11, -5.67046407e+11,
        -4.24536383e+11, -2.74411688e+11, -1.58206507e+11,
        -4.30553748e+10,  4.56966160e+10,  8.54628562e+10,
         8.75589374e+10,  7.80778984e+10,  5.61610816e+10,
         3.43744689e+10,  1.45553430e+10,  6.43628423e+09,
        -0.00000000e+00, -0.00000000e+00, -0.00000000e+00]])

```
---

Here's what I would do:
1. In `meteorology.py`
```python
    # zonal means
    da_v_zonal = ux_ipress.zonal_mean(lat=lat)
    da_v_zonal['plev'] = ux_ipress[plev_coordname]
    da_PS_zonal = uxds[surface_air_pressure_varname].zonal_mean(lat=lat)

    # Validate surface pressure before integration
    if da_PS_zonal is None:
        raise ValueError(
            "zonal_meridional_psi: surface air pressure variable could not be determined."
        )
    if np.isnan(da_PS_zonal).any():
        raise ValueError(
            "zonal_meridional_psi: zonal mean of surface air pressure for provided latitudes contains NaN values. "
            "Cannot compute pressure thickness."
        )
    if (uxds[surface_air_pressure_varname] <= 0).any():
        raise ValueError("zonal_meridional_psi: surface air pressure must be positive.")
```
This assumes that if any lat range returns a nan, we want to stop the calculation and fail. It also corrects a similar problem with `(da_PS_zonal <= 0).any()` -> `(uxds[surface_air_pressure_varname] <= 0).any()`

Ideally, we could do something like `da_PS_zonal.latitudes[np.isnan(da_PS_zonal)[0]].values` to tell the user what ranges might be causing the problem, but I don't know how to do that if there's multiple time steps robustly. I tried to make a test that used multiple time steps, and, uh, I got an error from inside uxarray. I'm going to make a separate comment about this.

2. In `test_meteorology.py`:
Edit the existing all nan test to reflect new error message and add a new test for weird lat ranges
```python
    def test_zonal_meridional_psi_raises_on_nan_zonal_mean(self, uxds_plev):
        """Test error when zonal mean contains NaN (achieved by passing all-NaN input)."""
        uxds_bad = uxds_plev.copy()
        uxds_bad["PS"][:] = np.nan

        with pytest.raises(ValueError) as errinfo:
            zonal_meridional_psi(uxds_bad, lat=self.lat)
        assert "contains NaN values" in str(errinfo.value)

    def test_zonal_meridional_psi_bad_lat_ranges(self, uxds_plev):
        # just one outside range of lats in uxds_plev
        with pytest.raises(ValueError) as errinfo:
            zonal_meridional_psi(uxds_plev, lat=self.lat.tolist() + [45])
        assert "contains NaN values" in str(errinfo.value)

        # barely inside range of lats in uxds_plev
        with pytest.raises(ValueError) as errinfo:
            zonal_meridional_psi(uxds_plev, lat=[30, 31, 32, 33, 34, 35, 36])
        assert "contains NaN values" in str(errinfo.value)

        # bigger range of lats outside of uxds_plev
        with pytest.raises(ValueError) as errinfo:
            zonal_meridional_psi(uxds_plev, lat=self.lat.tolist() + [45, 46, 47, 48])
        assert "contains NaN values" in str(errinfo.value)

        # all outside range of lats in uxds_plev
        with pytest.raises(ValueError) as errinfo:
            zonal_meridional_psi(uxds_plev, lat=[10, 11, 12, 13, 14, 15])
        assert "contains NaN values" in str(errinfo.value)
```

anissa111 (approved):

PR #776, opened 2025-10-15
pressure_at_hybrid_levels

### LINKED ISSUE #745: [💡]: Release `interpolation._pressure_from_hybrid` in the user API
[geocat.comp.interpolation._pressure_from_hybrid](https://geocat-comp.readthedocs.io/en/latest/internal_api/generated/geocat.comp.interpolation._pressure_from_hybrid.html) actually does what NCL's [pres_hybrid_ccm](https://www.ncl.ucar.edu/Document/Functions/Built-in/pres_hybrid_ccm.shtml) function did. 

I can't recall why we put it into the internal API, but please:
- Put it into the user API 
- Probably name it `pressure _at_hybrid_levels` or something similar (I don't know why we chose to say "from" at the time)
- Change places that used it so far accordingly 

### Is this a request for functionality that was previously in NCL?

Yes

### Additional context

> Hi @erogluorhan thanks for checking in. It's been a while since I was doing this kind of analysis, but reviewing the github issues and documentation pages, it seems like geocat covers this functionality through a combination of [geocat.comp.interpolation._pressure_from_hybrid](https://geocat-comp.readthedocs.io/en/latest/internal_api/generated/geocat.comp.interpolation._pressure_from_hybrid.html) and [geocat.comp.meteorology.dpres_plevel](https://geocat-comp.readthedocs.io/en/latest/user_api/generated/geocat.comp.meteorology.dpres_plev.html). So while NCL's [dpres_hybrid_ccm](https://www.ncl.ucar.edu/Document/Functions/Built-in/dpres_hybrid_ccm.shtml) is not exactly replicated, I think users could use a combination of those two geocat functions to do the same thing. 

 _Originally posted by @katiedagon in [#26](https://github.com/NCAR/geocat-comp/issues/26#issuecomment-3145709025)_

anissa111:
## PR Summary

This PR
- moves `_pressure_from_hybrid` into public API as `pressure_at_hybrid_levels`
- updates usages of previously internal function
- adds documentation for `pressure_at_hybrid_levels`
- makes function more robust by making sure multiplied dimensions are aligned
- adds testing (comparison with NCL's `pres_hybrid_ccm`)

Reviewer reference
- [new function page preview](https://geocat-comp--776.org.readthedocs.build/en/776/user_api/generated/geocat.comp.interpolation.pressure_at_hybrid_levels.html#geocat.comp.interpolation.pressure_at_hybrid_levels)
- new function inclusion on [API page preview](https://geocat-comp--776.org.readthedocs.build/en/776/user_api/index.html?readthedocs-diff=true&readthedocs-diff-chunk=1#interpolation)

Closes #745 

## PR Checklist
**General**

### DIFF
--- a/geocat/comp/interpolation.py
+++ b/geocat/comp/interpolation.py
@@ -389,7 +389,7 @@ def pressure_at_hybrid_levels(psfc, hya, hyb, p0=100000.0):
     elif isinstance(hya, xr.DataArray) and not isinstance(hyb, xr.DataArray):
         hyb = xr.DataArray(hyb, dims=hya.dims)
     elif not isinstance(hya, xr.DataArray) and isinstance(hyb, xr.DataArray):
-        hya = xr.DataArray(hyb, dims=hyb.dims)
+        hya = xr.DataArray(hya, dims=hyb.dims)
     # if both not DataArrays, convert both w/ dim 'lev'
     else:
         # convert to xarray for rest of checks and calculation


geocat/comp/interpolation.py
```diff
@@ -357,6 +348,49 @@ def _vertical_remap_extrap(
...
+    Parameters
+    ----------
+    psfc: :class:`xarray.DataArray`
+        A multi-dimensional array of surface pressures (Pa)
+
+    hya: :class:`xarray.DataArray`
+        An array of hybrid pressures (Pa)
+
+    hyb: :class:`xarray.DataArray`
+        An array of hybrid pressures (Pa)
```
kafitzgerald: I think these are coefficients to describe the hybrid levels.  The NCL docs have some language we could borrow: https://www.ncl.ucar.edu/Document/Functions/Built-in/pres_hybrid_ccm.shtml

geocat/comp/interpolation.py
```diff
@@ -357,6 +348,49 @@ def _vertical_remap_extrap(
...
+    --------
+    Related NCL Functions:
+    `pres_hybrid_ccm <https://www.ncl.ucar.edu/Document/Functions/Built-in/pres_hybrid_ccm.shtml>`__
+    """
+
+    # make sure hya and hyb dims align
+    # if not, assign hya dim to hyb
+    if isinstance(hya, xr.DataArray) and isinstance(hyb, xr.DataArray):
+        if hya.dims != hyb.dims:
+            hyb = hyb.rename({hyb.dims[0]: hya.dims[0]})
```
kafitzgerald: Do we want to do this for folks?

kafitzgerald (commented):
Aside from the docstring and maybe the dim renaming (open to being persuaded on this), looks good to me.

geocat/comp/interpolation.py
anissa111: Yeah, I don't know.

My thought was that if the dim names don't match and they're trying to do this calculation, they likely intended for them to match. 

I'd be okay with doing alternatives like a temp internal var or restoring the original dim names afterward.

anissa111: Also, testing locally, I added print statements to look at the dims before and after the function call and they appear to be unchanged 

screenshot of tests with

```
('ncl0',)
('ncl1',)
('ncl0',)
('ncl1',)
```

geocat/comp/interpolation.py
anissa111: Ooh good, yeah I'll do that!

Edit: done!

geocat/comp/interpolation.py
kafitzgerald: Yeah, beyond not loving side effects - I worry a bit that if this case arises it might be an indicator that the user is doing something unintended (most the time the coefficients should be coming from the same file w/ the same dims) and that changing the dims w/out notifying them could make for a confusing bug somewhere down the line.

Xarray requiring alignment probably shouldn't be a surprise to folks, but maybe it is and I don't know how this is handled elsewhere in comp.

A temp var sounds reasonable.

kafitzgerald: Never mind, it looks like this doesn't actually cause a side effect (sorry for the false alarm).

And in that case I care quite a bit less.  Happy to have the fix things for folks vs warn them, etc. convo later.  I think this is pretty unlikely to come up anyway.

anissa111: I tried adding a temp var to hold the original dims and swapped them back afterwards (https://github.com/NCAR/geocat-comp/pull/776/commits/2df2c1a4a2b00023808d5765f18ba4990d69815c), but my IDE was fairly certain that the swapped back version would be unused. I can't think of a scenario where that would happen, and I suppose I could do something like import the global version of the variable to check against, but I think that may be more confusing or introduce more bugs than leaving it as is.

I did increase the robustness of the dimensionality checking and testing, though, so hopefully if anything unexpected gets passed in, it'll be easier to debug.

geocat/comp/interpolation.py
```diff
@@ -381,10 +381,21 @@ def pressure_at_hybrid_levels(psfc, hya, hyb, p0=100000.0):
...
     # make sure hya and hyb dims align
-    # if not, assign hya dim to hyb
     if isinstance(hya, xr.DataArray) and isinstance(hyb, xr.DataArray):
+        # if not, assign hya dim to hyb
         if hya.dims != hyb.dims:
             hyb = hyb.rename({hyb.dims[0]: hya.dims[0]})
+    else:
+        # convert to numpy for rest of checks and calculation
+        hya = xr.DataArray(hya, dims={'levels'})
+        hyb = xr.DataArray(hyb, dims={'levels'})
```
kafitzgerald: What happens if only one is a DataArray though?  It's unlikely, but it will break this.

If this isn't something we account for elsewhere though - don't worry about it.

anissa111: Ooh good point. I think it's okay, though.

I added mixed input types: 
```python
def test_pressure_at_hybrid_levels(self, p_out, ds_out):
    # run not officially supported input types w/ hyam and hybm as lists and mixed
    pm = pressure_at_hybrid_levels(
        self.ps[0, :7, :7], p_out.hyam.values.tolist(), p_out.hybm.values.tolist()
    )
    nt.assert_allclose(pm, p_out.pm, rtol=1e-6)
    pm = pressure_at_hybrid_levels(self.ps[0, :7, :7], p_out.hyam.values.tolist(), p_out.hybm)
    nt.assert_allclose(pm, p_out.pm, rtol=1e-6)
    pm = pressure_at_hybrid_levels(self.ps[0, :7, :7], p_out.hyam, p_out.hybm.values.tolist())
    nt.assert_allclose(pm, p_out.pm, rtol=1e-6)

    # mismatched dim names
    pm = pressure_at_hybrid_levels(self.ps[0, :7, :7], p_out.hyam, p_out.hybm)
    nt.assert_allclose(pm, p_out.pm, rtol=1e-6)
```

and it's still passes. It still runs in 4ms, so I'll push those lines, too

kafitzgerald: Interesting.

If you try it with this file, which has `lev` set as a coordinate rather than just a dimension it'll fail with a CoordinateValidationError.
`ds = xr.open_dataset(gcd.get('netcdf_files/ccsm35.h0.0021-01.demo.nc'))`

anissa111: ~We're already using that file in the pytest module, so I moved it to a module-level fixture~

~and added a test that just tries to run `pressure_at_hybrid_levels` on it:~

~Locally, this seems to be working for me. I'll push it (edit: https://github.com/NCAR/geocat-comp/pull/776/commits/cc3895fab3e5fddf6d9eb2615a2b47c1dcdd82ba) and make sure it's not a fluke.~

~How did yours look when you got the CoordinateValidationError?~

Edit: Putting this here for future reference, the CoordinateValidationError pops up when there are dims on either `hyam` or `hybm`, while the other is not a DataArray. 

screenshot of test with described scenario

I've added this to deal with that input scenario:
```python
# check if only one is a DataArray, convert the other
elif isinstance(hya, xr.DataArray) and not isinstance(hyb, xr.DataArray):
    hyb = xr.DataArray(hyb, dims=hya.dims)
elif not isinstance(hya, xr.DataArray) and isinstance(hyb, xr.DataArray):
    hya = xr.DataArray(hya, dims=hyb.dims)
```

PR #821, opened 2026-03-02
[for testing purposes] asv action regression failure?

anissa111:
Feel free to ignore this PR, I'm trying to test the ASV PR comparison action after some unexpected failures on #773.

My theory is that the asv compare command may be returning some sort of "failure" if a regression is detected, so my plan is to:
- run benchmarks on an empty commit, essentially comparing main against main
- add a delay to a function to artificially introduce a regression
- see if the asv pr comparison action starts having a problem only after a regression is artificially introduced


anissa111:
Yeah, the `asv compare` is outputting something that reads as a failure to the action.

PR #851, opened 2026-05-04
Ignore AMS journal links in linkcheck

kafitzgerald:
## PR Summary

Links to AMS journal articles don't work with our link checker.  This adds them to the `linkcheck_ignore` configuration.


kafitzgerald:
I skipped the issue and release notes addition for this since it's super minor and not at all user facing, but could be convinced to add those if others find value in that.

cyschneck (approved):

